PR #67, opened 2016-11-21
Use the project.version variable, instead of another variable in pom.xml

kinow:
The [project.version variable](https://maven.apache.org/guides/introduction/introduction-to-the-pom.html#Project_Interpolation_and_Variables) can be used to retrieve the current version in pom.xml.

With that, it won't be necessary to keep both cantaloupe.version and the pom.xml version in sync, preventing issues, and simplifying the build pipeline.

Tested locally, get the Cantaloupe-3.3-SNAPSHOT.zip in my target folder, just as before the change.

Bruno


kinow:
Resubmitting as #68 with the correct commit e-mail.

PR #70, opened 2016-11-24
Fix fragment link

kinow:
Fixing a link to a fragment that has an extra #


adolski:
Thanks for fixing all of these!

PR #75, opened 2016-12-01
Add usage method, and print the usage when no config was provided, instead of throwing NullPointerException

kinow:
The first time I started the application in my Eclipse workspace I got a NPE, and didn't understand if I was doing some mistake, or if there was a bug in the develop branch.

The output also didn't include the usage, but after reading the code I noticed that I had to provide a configuration file (I should have read the manual actually, I know :-) ).

This pull request adds a usage method, and checks if the configuration object was not created, and if so, prints the usage and exists.

It also prints the usage for the other cases like when there is no file, or instead of a file the user provided a directory.

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/StandaloneEntry.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/StandaloneEntry.java
@@ -36,27 +36,27 @@ public class StandaloneEntry {
     public static void main(String[] args) throws Exception {
         final Configuration config = ConfigurationFactory.getInstance();
         if (config == null) {
-            usage();
+            printUsage();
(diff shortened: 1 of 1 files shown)


adolski:
Yeah, it was supposed to print that message, but it looks like I broke it awhile back. Thanks for fixing it. Would you mind renaming `usage()` to `printUsage()`?

kinow:
Don't mention. I'm having fun slowly learning about iiif, reading through Cantaloupe code base, and toying around Jenkins, Open Seadragon, etc :)

>Would you mind renaming usage() to printUsage()?

Not at all. Just pushed a new commit :+1:

PR #102, opened 2017-04-14
Update Java2DUtil.java

Algae666:
Allow config for background color when flattening PNG to JPG

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/processor/Java2DUtil.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/processor/Java2DUtil.java
@@ -55,8 +55,6 @@ public abstract class Java2DUtil {
 
     private static Logger logger = LoggerFactory.getLogger(Java2DUtil.class);
 
-    static final String FLATTEN_COLOR_CONFIG_KEY = "processor.flatten_color";
-
     /**
      * See the inline documentation in scaleImage() for a rationale for
      * choosing this.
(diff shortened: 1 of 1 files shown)


src/main/java/edu/illinois/library/cantaloupe/processor/Java2DUtil.java
```diff
@@ -554,6 +558,13 @@ public static BufferedImage removeAlpha(final BufferedImage inImage) {
             outImage = new BufferedImage(inImage.getWidth(),
                     inImage.getHeight(), newType);
             Graphics2D g = outImage.createGraphics();
+            try {
+                String bgColor=Configuration.getInstance().getString(FLATTEN_COLOR_CONFIG_KEY);
```
adolski: Can we use `Processor.BACKGROUND_COLOR_CONFIG_KEY` instead?

src/main/java/edu/illinois/library/cantaloupe/processor/Java2DUtil.java
```diff
@@ -554,6 +558,13 @@ public static BufferedImage removeAlpha(final BufferedImage inImage) {
             outImage = new BufferedImage(inImage.getWidth(),
                     inImage.getHeight(), newType);
             Graphics2D g = outImage.createGraphics();
+            try {
```
adolski: I think this try/catch is unnecessary

src/main/java/edu/illinois/library/cantaloupe/processor/Java2DUtil.java
Algae666: I used a different config, because it looks like BACKGROUND_COLOR_CONFIG_KEY is specifically used for rotation. I don't have a use case for a different background color for flattening alpha channels vs rotation, but I didn't know if someone else may. I'm totally good with using the existing config. Also, just put in the try/catch because if someone uses an invalid color in the configs, it could throw an exception, but that's a style thing. I'm ok without it.

adolski:
Great, thanks for contributing this. There are some additional changes that will have to be made to make this integrate with caching, and I will take care of those soon.

PR #118, opened 2017-06-13
Add flexibility in AWSS3Resolver

### LINKED ISSUE #103: Added flexibility in AWSS3Resolver
Hi there. We'd like to use the AWSS3Resolver but in order to work in our deployment topology we've had to make 2 changes:

1. Use a `CredentialProviderChain` to take the credentials from other environment parameters if not specified in the properties file
2. Allow the `get_s3_object_key` delegate method to return an array with the bucket name and key for cases where the bucket can't be hard coded in the properties file

Would you be open to considering a pull request with these changes?

jweisman:
Hi there. As we discussed, these changes allow the key, secret, and region to be taken from the environment if not provided in the config file. I also added a hash as a valid response from AmazonS3Resolver.get_object_key. This is meant to fix #103.

I would love to get your feedback. Thanks.


src/main/java/edu/illinois/library/cantaloupe/resolver/AmazonS3Resolver.java
```diff
@@ -90,9 +117,23 @@ private S3Object getObject() throws IOException {
...
+        final String bucketName;
+        final String objectKey;
+        if (object instanceof RubyHash) {
+        	Map<String, Object> map = convertToHashMap((RubyHash)object);
+        	if (map.containsKey("bucket") && map.containsKey("key")) {
+	        	bucketName = map.get("bucket").toString();
+	        	objectKey = map.get("key").toString();
+        	} else {
+        		logger.error("Hash does not include bucket and key");
+        		throw new IOException();
```
jweisman: Sorry- didn't follow this one?

jweisman:
Thanks @adolski for reviewing and for your comments. All seem reasonable to me and I added a commit with the changes. 

I left the RubyHash reference only because I felt it was more readable, but let me know if you feel strongly.

adolski:
Thanks @jweisman! Sorry for the slowness, I've been busy on another project.

jweisman:
Thanks @adolski, and thanks for all your work on this great project.

Issue #130, opened 2017-07-24
Add instructions for running the tests


kinow:
Hi,

I will be presenting in a local conference in a few days about IIIF, and using Cantaloupe as Image server for experimentations and demo.

I have the latest version from the develop branch, but not all tests pass. I remember running the tests some time ago, and got a similar error. Are there any instructions on how to run the tests?

Normally, when I find a Java Maven project in GitHub, and `mvn clean test` does not work, I look for the Travis CI file in the project repository.

Maybe we could add instructions for building and running - if these instructions are not already somewhere - the project, and also include Travis-CI for the branches and pull requests?

I'm also getting very intriguing results when running the tests. This is the last commit I have in my working copy:

```
commit a4c874847894b2936aa81f0350349419890b7f5a
Author: Alex Dolski <alexd@n-ary.net>
Date:   Tue Jul 18 14:18:31 2017 -0500
```

Running `mvn clean test`, I got:

```
Tests run: 1341, Failures: 8, Errors: 156, Skipped: 21
```

After some minutes, running `mvn clean test` again, I got:

```
Tests run: 1341, Failures: 8, Errors: 135, Skipped: 21
```

The number of tests run, failures, and skipped is consistent. But the number of errors is not. I didn't investigate why, as I'm working to run Cantaloupe within Eclipse and also finishing slides. But if it's a known issue, I could try to dig deeper and perhaps help with this if necessary :-)

Ta
Bruno

adolski:
I can certainly add a testing section to the readme. The basic idea is:

1. Install dependencies: GraphicsMagick, ImageMagick, FFmpeg, Kakadu, OpenJPEG, Redis, and Firefox+geckodriver (needed by Selenium for ControlPanelTest). Whichever of these are missing will cause the corresponding tests to fail, so if Kakadu is missing, KakaduProcessorTest will fail, etc.

2. Copy `test.properties.sample` to `test.properties` and fill in whatever info you have -- again, anything missing will cause failures.

I don't run the tests as often as I should, so I'm not surprised that there are some failures. There should only a handful of consistent ones, though. I just ran and got 19 errors & 3 failures, and there should be less than that.

I can reproduce the inconsistent results across runs. Often I run individual package tests from IntelliJ, and they run fine. I don't know what the problem is -- my only guess would be state being carried over across the tests, but I don't know where it might be occurring. I'd love to get them all consistently passing.

adolski:
[...]

PR #136, opened 2017-08-02
Use a test profile, and fix some tests

kinow:
Right now I have one test failing. This PR includes a `nodeps` profile as discussed in #130, for running the unit tests that do not require external accounts in AWS/Azure/etc, nor certain dependencies such as Kakadu or Selenium.

There are also several tests that use font family, and not family name. The problem with that is that the value changes quite a bit in Linux/Windows. So this PR also addresses this problem, by using the font name.

The Helvetica font is not installed in Ubuntu, or at least is not easily installed. Tried the `ms-mscorefonts-installer` but didn't get the Helvetica font, so changed by Arial.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -526,4 +526,33 @@
 
     </plugins>
   </build>
+
+  <profiles>
+    <profile>
+      <id>nodeps</id>
+      <activation>
+        <activeByDefault>false</activeByDefault>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <version>2.19.1</version>
+            <configuration>
+              <runOrder>random</runOrder>
+              <reuseForks>false</reuseForks>
+              <excludes>
+                <exclude>Azure*Test</exclude>
+                <exclude>Amazon*Test</exclude>
+                <exclude>Kakadu*Test</exclude>
+                <exclude>Redis*Test</exclude>
+                <exclude>ControlPanelTest</exclude>
+              </excludes>
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+  </profiles>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -543,8 +543,37 @@
               <runOrder>random</runOrder>
               <reuseForks>false</reuseForks>
               <excludes>
-                <exclude>Azure*Test</exclude>
+                <exclude>AmazonS3*Test</exclude>
+                <exclude>AzureStorage*Test</exclude>
+                <exclude>FfmpegProcessorTest</exclude>
+                <exclude>GraphicsMagickProcessorTest</exclude>
+                <exclude>ImageMagickProcessorTest</exclude>
+                <exclude>KakaduProcessorTest</exclude>
+                <exclude>OpenJpegProcessorTest</exclude>
+                <exclude>RedisCacheTest</exclude>
+              </excludes>
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+    <profile>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -526,4 +526,33 @@
...
+            <version>2.19.1</version>
+            <configuration>
+              <runOrder>random</runOrder>
+              <reuseForks>false</reuseForks>
+              <excludes>
+                <exclude>Azure*Test</exclude>
+                <exclude>Amazon*Test</exclude>
+                <exclude>Kakadu*Test</exclude>
+                <exclude>Redis*Test</exclude>
+                <exclude>ControlPanelTest</exclude>
```
adolski: I think the complete list would be:

    <exclude>AmazonS3*Test</exclude>
    <exclude>AzureStorage*Test</exclude>
    <exclude>FfmpegProcessorTest</exclude>
    <exclude>GraphicsMagickProcessorTest</exclude>
    <exclude>ImageMagickProcessorTest</exclude>
    <exclude>KakaduProcessorTest</exclude>
    <exclude>OpenJpegProcessorTest</exclude>
    <exclude>RedisCacheTest</exclude>

I excluded ControlPanelTest as I switched it to use HtmlUnit instead of Firefox+geckodriver, so it no longer has external dependencies (although it's still not working and marked as `@Ignore`).

kinow:
Hi @adolski

>Do you want to update the testing section of the readme as well? No obligation, I'm happy to do it.

Oh, good idea! I can send a PR in a few hours with some text for review.

>I think the complete list would be:

Will try with these values when I arrive at work, and update this PR. Thanks!!!

pom.xml
kinow: Oh, so looking at the complete list, I think we can actually keep Ffmpeg, GraphicsMagick, ImageMagick, and OpenJpeg? Or maybe have another profile to include these...

Looks like we have 1) minimum of tests that can be executed in a blank box with Java + Maven, 2) a few more tests that require installation of open source / free tools & libraries (e.g. ffmpeg, graphicsmagick, imagemagick, openjpeg), 3) other tests that require licences or paid tools (e.g. kakadu, amazon, azure).

For the Travis CI tests, we can include 1) + 2). For 2) we'd need just to use a Ubuntu node, and [install some packages](https://docs.travis-ci.com/user/installing-dependencies/).

So do you think we should keep all tests executing by default, have a `nodeps` profile for developers and users to try (with your list above), and another profile for continuous integration (maybe `ci` ?) ?

adolski: Sorry, I was thinking of everything with external dependencies, but yeah, that's the complete list of everything with non-free dependencies.

> So do you think we should keep all tests executing by default, have a nodeps profile for developers and users to try (with your list above), and another profile for continuous integration (maybe ci ?) ?

How about: `freedeps` (used with CI), and then either `nodeps` or `alldeps`?

I'm neutral -- whatever you want to go with.

kinow:
Rebase'd commit against develop branch. Added `freedeps` profile. Ordered alphabetically the `nodeps` dependencies. Updated README with instructions for running tests.

Right now, here's what I get in my environment:

`mvn clean test -Pfreedeps`

Failed tests: 
  StandaloneEntryTest.testMainWithPurgeExpiredCacheArg:410 expected:<1> but was:<0>
  Java2DUtilTest.testApplyOverlayWithStringOverlay:194

Tests run: 1294, Failures: 2, Errors: 0, Skipped: 21

`mvn clean test -Pnodeps`

Failed tests: 
  StandaloneEntryTest.testMainWithPurgeExpiredCacheArg:410 expected:<1> but was:<0>
  Java2DUtilTest.testApplyOverlayWithStringOverlay:194

Tests run: 1222, Failures: 2, Errors: 0, Skipped: 20

pom.xml
kinow: Done! Using `freedeps` and `nodeps`.

PR #141, opened 2017-08-07
Add new context parameter to resolver delegates.

### LINKED ISSUE #137: Authorization information to resolver delegate
First I just have to say Cantaloupe is an awesome image server, and I really appreciate all the work that has gone into it. 

I'm having a bit of a problem with my authorization workflow. I am implementing an authorization delegate which verifies a username and password that is passed in with the request headers. So far this is all okay. 

However this username and password information also needs to be included in the HTTP request to the backend using the `HttpResolver`. I was hoping I would be able to sort out this information in the `ScriptLookupStrategy ` delegate, but it doesn't seem to have any access to headers.

My question is: Is there any way that I am missing to pass information from the authorization delegate to the HttpResolver delegate, or is there any way to get access to headers in the HttpResolver delegate? 

If the answer to the above questions is no, would there be any interest in a PR that implemented functionality to pass some sort of an opaque auth object between the authorization delegate and the resolver delegates.

Thanks!

jonathangreen:
Resolves #137 

I believe I've implemented everything you mentioned in that issue. Let me know if I missed something.

### DIFF
--- a/delegates.rb.sample
+++ b/delegates.rb.sample
@@ -87,10 +87,11 @@ module Cantaloupe
 
     ##
     # @param identifier [String] Image identifier
+    # @param context [Hash] Context for this request
     # @return [String,nil] Absolute pathname of the image corresponding to the
     #                      given identifier, or nil if not found.
     #
-    def self.get_pathname(identifier)
+    def self.get_pathname(identifier, context)
     end
 
   end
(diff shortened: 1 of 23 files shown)


adolski:
Thanks very much for this, @jonathangreen. I think it's almost ready to merge. Just two things:

1. Several of the resolver tests are failing. I believe `AbstractResolver.context` is null when testing, so the resolvers' calls to `context.asMap()` throw a NullPointerException.

2. Also, could you edit `website/upgrade.html` to tell people to add the `context` arg to their resolver delegate methods?

jonathangreen:
@adolski Just pushed a few commits to fix those issues. I don't have credentials to test all the services with, so I can't actually run all of the resolver tests, but what I can run is working for me now. Let me know if you see any other issues with tests failing.

adolski:
Excellent, thanks.

Issue #140, opened 2017-08-07
Standardized tests for Cantaloupe implementations


scossu:
In order to effectively measure the performance and features of Cantaloupe (or, for what matters, any image server) it would be very valuable to have a test suite that can easily be set up in any existing Cantaloupe implementation. 

The goal is to provide the widest possible terms of comparison between different IIIF server implementations by performing identical and predictable requests on different architectures. 

N.B. This ticket only focuses on Image API performance. 

Characteristics of this test suite should be: 

1. Portability: written in a widespread programming language with few or no dependencies for an average server environment (e.g. PHP, Pyhon, Ruby)
2. Diversity: as many diverse requests as possible should be performed while maintaining the code reasonably uncomplicated
3. Replicability: non-random tests should yield similar results in two identical systems (the difference only being caused by traffic and load not related to the tests)

Some sample images should be provided—at least one large (>4096 px per side) pyramidal TIFF and one JP2 of the same size, as well as a medium-size (~1920 px) pair. All images should be derived from the same source. 

Example tests could be: 

- Retrieve a variety of tiles and other derivatives from the same image; move on to a different format, then to a different base size
- Same as above with 1, 2, 4, 8, 16, 32, 64... concurrent connections
- Compare the above by retrieving tiles in an ordered fashion (i.e. many requests on the same tile at the same time) vs. random order (many different tiles requested at the same time) 
- Isolated tests for region, scaling and rotation

adolski:
Hi @scossu,

This sounds like a legitimate and useful tool. I have to say, though, that I don't want to incorporate it into Cantaloupe, either built-in or as an separate tool. Since it could work with any compliant Image API implementation, there would be no reason to couple it to this one, and so it would be better off in its own repo where it's easier to independently release etc. It's basically just a different project.

PR #144, opened 2017-08-18
[WIP] Add performance benchmark tests with JMH

kinow:
This is related to #140. It is not as comprehensive as suggested in the issue. But this pull request contains an example of microbenchmarking with [JMH](http://openjdk.java.net/projects/code-tools/jmh/).

The two benchmark tests included simply extend existing TIFF reader/write tests. Ideally, I think we would like to map a few operations that are considered important and worth keeping an eye every now and then.

Users & developers are then able to run the test in their environment and compare with other results by other users.

This can be useful when choosing hardware for a Cantaloupe server, investigating slowness, and checking for regressions in the code (i.e. run the JMH tests against tag 1.0, and later compare with results from tag 2.0 and see what was the difference).

It is possible to run these tests in Travis CI too, though it requires further investigation if worth, as it may take a while - I suspect running occasionally before releases with important changes should suffice.

Running the tests:

```
$ mvn clean test -Pbenchmark
# the test is just the phase. The normal tests are being ignored, see the profile settings for more
# and it may take several minutes depending on hardware and tests
# then eventually you get something like...
# Run complete. Total time: 00:01:54

Benchmark                                                                   Mode  Cnt     Score     Error  Units
TIFFImageReaderPerformance.testGetCompression                               avgt    5  3478.766 ± 386.671  us/op
TIFFImageReaderPerformance.testGetMetadata                                  avgt    5   384.149 ±  32.546  us/op
TIFFImageReaderPerformance.testReadWithMonoResolutionImageAndNoScaleFactor  avgt    5   331.710 ±  21.164  us/op
TIFFImageWriterPerformance.testWriteWithBufferedImage                       avgt    5  1287.574 ± 123.852  us/op
TIFFImageWriterPerformance.testWriteWithBufferedImageAndExifMetadata        avgt    5  2486.737 ± 295.360  us/op
TIFFImageWriterPerformance.testWriteWithBufferedImageAndIptcMetadata        avgt    5  2657.941 ± 270.055  us/op
TIFFImageWriterPerformance.testWriteWithBufferedImageAndXmpMetadata         avgt    5  2421.456 ± 297.486  us/op
TIFFImageWriterPerformance.testWriteWithPlanarImage                         avgt    5  1389.420 ±  99.007  us/op
TIFFImageWriterPerformance.testWriteWithPlanarImageAndExifMetadata          avgt    5  2553.279 ± 335.514  us/op
TIFFImageWriterPerformance.testWriteWithPlanarImageAndIptcMetadata          avgt    5  2323.278 ± 134.614  us/op
TIFFImageWriterPerformance.testWriteWithPlanarImageAndXmpMetadata           avgt    5  2346.666 ± 215.869  us/op

Benchmark result is saved to target/jmh-result.edu.illinois.json
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 02:02 min
[INFO] Finished at: 2017-08-18T14:44:29+12:00
[INFO] Final Memory: 37M/402M
[INFO] ------------------------------------------------------------------------
```

In this case the output is in [average time](http://hg.openjdk.java.net/code-tools/jmh/file/6cc1450c6a0f/jmh-core/src/main/java/org/openjdk/jmh/results/AverageTimeResult.java#l39). So the `testGetCompression` was executed 5 times, and the average time was 3478.766 micro-seconds (3.4 milliseconds, under 0.004 seconds :-)).

Marking as DISCUSSION as it would require a confirmation it is desirable to have it in the project. I read the comment in #140, and I agree about the scope if the tests are supposed to guarantee backward compatibility, compatibility with other tools, etc. But these tests would be simply for micro benchmarking. In other words, we could use it if someone suggested that in the current release, a certain operation has degraded in performance.

And it is also for DISCUSSION because the current examples are simply to illustrate how it could be implemented. I agree with original request in #140 that we would want to have some good input files, and also specific operations that are considered important.

Example of projects using JMH, from where I adapted the configuration for Cantaloupe:

* [Apache Commons CSV](https://github.com/apache/commons-csv/blob/431f8236e89acd1bd3c678fad727f2d585226be8/pom.xml#L398)
* [Apache Commons RNG (random number generator)](https://github.com/apache/commons-rng/tree/b8fa4766c710d9e57fb866180dd17180d192bb70/commons-rng-jmh)

Cheers
Bruno

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -214,10 +214,24 @@
       <artifactId>pdfbox</artifactId>
       <version>2.0.4</version>
     </dependency>
+    <!-- Unit tests -->
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <version>4.12</version>
+      <scope>test</scope>
+    </dependency>
+    <!-- Performance tests -->
+    <dependency>
+      <groupId>org.openjdk.jmh</groupId>
+      <artifactId>jmh-core</artifactId>
+      <version>1.19</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.openjdk.jmh</groupId>
+      <artifactId>jmh-generator-annprocess</artifactId>
+      <version>1.19</version>
       <scope>test</scope>
     </dependency>
     <!-- Assists in parsing XMP metadata in RDF/XML -->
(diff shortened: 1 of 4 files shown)


adolski:
Sorry for the delay on this -- this is really cool! My understanding of #140 is that it calls for a tool to measure performance of any image server against the IIIF Image API, i.e. it's not coupled to any particular implementation. This is different in that it would test primarily internal API. (Though it could test the endpoints too.) It will put some firmer measurements on things that I've always assumed but never verified.

I've been waiting for a free chunk of time to devote to this, but I haven't found one, so I think I'm going to just leave it here and push to it as time permits. (Is that OK?) Eventually, I would like to cover a lot more.

kinow:
>I've been waiting for a free chunk of time to devote to this, but I haven't found one, so I think I'm going to just leave it here and push to it as time permits. (Is that OK?) Eventually, I would like to cover a lot more.

OK for me. I'm busy with other projects, and still trying to find some time to have fun with Apache Commons Imaging + Cantaloupe, as well as tile caching. So feel free to take your time on this one :-)

Thanks!
B

adolski:
Thanks again for this, @kinow. I'm planning on expanding it greatly and will track that in #164.

kinow:
Excellent news @adolski ! Will keep an eye on that issue for more exciting news!

ps: keep an eye on commons-imaging / sanselan. Starting to get more familiar with its code base ;-)

PR #146, opened 2017-08-18
Fix minor import issues and generics

kinow:
As per the title, this PR fix minor import and generic issues.

Cheers
Bruno


kinow:
Realized I have been commenting that the tests pass... but maybe we could enable Travis to test pull requests too? Looks like it's not monitoring them right now.

PR #152, opened 2017-09-20
Make improvements to the IIIF sizes metadata

cbeer:
This PR does two things, and I'm happy if you want to take one, both, or neither.

- Makes the IIIF minimum image size configurable (was hardcoded at 64px; we'd like to raise that locally)
- Adds the full size image to the list of `sizes` (I can get into our rationale, but, basically, we want our level 0-compatible clients to also have access to the full size image without needing to make inferences based on the profile/width/height/maxwidth/maxheight properties)


adolski:
Thanks very much @cbeer!

PR #154, opened 2017-09-22
Guard against cache directory creation race conditions

cbeer:
I believe we're seeing a race condition in the FilesystemCache when we try to use a shared cache across multiple Cantaloupe servers. I believe a patch like this might address the problem?


adolski:
Hi @cbeer,

Thanks for reporting this, and sorry for the delay. I see the problem, but I decided to fix it a little differently. I made the commit against the `release/3.3` branch, and merged into `develop`. It would be great if you're able to take it out for a test drive.

cbeer:
👍 what you've done looks good to me (and I figured it was easier to point out the problem in code than try to describe it). We'll certainly give this a try next week.

cbeer:
@adolski: we rolled this out earlier this week, and your fix seems to have fixed our problem, thanks.

PR #181, opened 2017-12-05
Ensure that the source cache is also purged, and refactor a bit.

### LINKED ISSUE #180: FilesystemCache source cache expiration
Thanks for the great project. It's been very useful for us.

The `FilesystemCache` does not perform purging of the source cache based on expiration. I noticed that this behavior has been changed in devel. Would you consider a patch to address this for 3.3.5? Since that code has been re-written fairly extensively, I'd probably elect to just keep it simple and copy the block seen here, but with the appropriate changes for the source cache:

https://github.com/medusa-project/cantaloupe/blob/750deaf17c9c2c44c12e2d1f77a3c782e1ce8307/src/main/java/edu/illinois/library/cantaloupe/cache/FilesystemCache.java#L869

jhosteny:
This patch fixes #180. Note that I took the opportunity to re-factor the code a bit as well, rather than just copy the block, since it was a third instance of the cleaner.

FWIW, we ran this today in production with a cache that had over 12 GB in the source cache, a TTL of 3600 seconds and cleaner running every 1800 seconds.

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/cache/FilesystemCache.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/cache/FilesystemCache.java
@@ -844,44 +844,34 @@ class FilesystemCache implements SourceCache, DerivativeCache {
                 }
             }
         }
-        logger.info("purgeExpired(): purging...");
 
         try {
             globalPurgeInProgress.set(true);
-            final String imagePathname = rootDerivativeImagePathname();
-            final String infoPathname = rootInfoPathname();
 
-            long imageCount = 0;
-            final File imageDir = new File(imagePathname);
-            Iterator<File> it = FileUtils.iterateFiles(imageDir, null, true);
-            while (it.hasNext()) {
-                final File file = it.next();
-                if (isExpired(file)) {
+            final String[] pathnamesToPurge = {
+                rootSourceImagePathname(),
+                rootDerivativeImagePathname(),
+                rootInfoPathname() };
+            for (String pathname : pathnamesToPurge) {
+                logger.info("purgeExpired(): purging {}...", pathname);
+                long fileCount = 0;
+                final File fileDir = new File(pathname);
+                Iterator<File> it = FileUtils.iterateFiles(fileDir, null, true);
+                while (it.hasNext()) {
                     try {
-                        FileUtils.forceDelete(file);
-                        imageCount++;
+                        final File file = it.next();
+                        if (isExpired(file)) {
+                            FileUtils.forceDelete(file);
+                            fileCount++;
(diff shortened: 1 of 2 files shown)


src/main/java/edu/illinois/library/cantaloupe/cache/FilesystemCache.java
```diff
@@ -844,44 +844,34 @@ public void purgeExpired() throws CacheException {
...
+		Iterator<File> it = FileUtils.iterateFiles(fileDir, null, true);
+		while (it.hasNext()) {
+		    try {
+			final File file = it.next();
+			if (isExpired(file)) {
+			    FileUtils.forceDelete(file);
+			    fileCount++;
+			}
+		    } catch (FileNotFoundException e) {
+			logger.debug(e.getMessage(), e);
```
jhosteny: I made this a debug statement, since it seemed to be not to uncommon. I saw about 20 instances of this over the course of a 4 hour run which populated the source cache with ~12 GB of data. It typically seemed to occur as a result of a race with https://github.com/BossaNova/cantaloupe/blob/5fc0373a3499ddb3398ee0893f7d19e79df67d55/src/main/java/edu/illinois/library/cantaloupe/cache/FilesystemCache.java#L234

src/main/java/edu/illinois/library/cantaloupe/cache/FilesystemCache.java
```diff
@@ -844,44 +844,34 @@ public void purgeExpired() throws CacheException {
...
-                    }
-                }
-            }
-            logger.info("purgeExpired(): purged {} expired image(s) and {} " +
-                    "expired infos(s)", imageCount, infoCount);
+            final String[] pathnamesToPurge = {
+		rootSourceImagePathname(),
+		rootDerivativeImagePathname(),
+		rootInfoPathname() };
+            for (String pathname : pathnamesToPurge) {
```
jhosteny: I refactored a bit to be a bit more like this: https://github.com/BossaNova/cantaloupe/blob/5fc0373a3499ddb3398ee0893f7d19e79df67d55/src/main/java/edu/illinois/library/cantaloupe/cache/FilesystemCache.java#L447

adolski:
Thanks for taking care of this, and for the detailed explanation of what you did. The changes look good to me. One thing, though: would you mind changing the tab indents to 4 spaces?

jhosteny:
@adolski argh! I'm sorry about that. Should be fixed now.

adolski:
Thanks @jhosteny!

jhosteny:
Thanks, appreciate the project!

PR #188, opened 2017-12-28
Add support for storage URI and SAS Token directly in URL.

ddisciascio:
Would like to have this considered to add support to the Azure resolver to directly connect to blob store to get images using Storage URI path and Sas tokens:
https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#resource-uri-syntax
https://docs.microsoft.com/en-us/azure/storage/common/storage-dotnet-shared-access-signature-part-1#how-a-shared-access-signature-works

### DIFF
--- a/cantaloupe.properties.sample
+++ b/cantaloupe.properties.sample
@@ -198,12 +198,15 @@ AmazonS3Resolver.lookup_strategy = BasicLookupStrategy
 #----------------------------------------
 
 # !! Name of your Azure account.
+# !! Leave blank if using URI with a SAS token in your object key.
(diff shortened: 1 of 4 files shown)


adolski:
Hi @ddisciascio,

Thanks for offering this. I can't accept it into `release/3.3` as the release branches are for bugfixes only, but I would be glad to take it into `develop`, which is tracking 3.5. There have been some changes to AzureStorageResolver since 3.3 and I hope porting this over won't be too much work.

PR #189, opened 2018-01-05
Support SAS + storage uri direct to azure blob store.

ddisciascio:


adolski:
Thanks @ddisciascio !

Issue #190, opened 2018-01-09
OpenJpegProcessor operating on low bit-depth images


adam-vessey:
There seems to be a [known issue with bit depths, due to the use of BMP intermediates](https://github.com/medusa-project/cantaloupe/blob/ffbfb5d41f80fd2b8beec155567938912048668d/src/main/java/edu/illinois/library/cantaloupe/processor/OpenJpegProcessor.java#L58-L60); however, it seems to describe the opposite situation. When testing in the wild, we have come across grayscale JPEG 2000 images with a bit-depth less than 8, which [OpenJPEG does not support with BMP](https://github.com/uclouvain/openjpeg/blob/9d0d1a0128c7a9324c9a935b2ded01deb4d660d4/src/bin/jp2/convertbmp.c#L866-L870)... this error makes its way into log files:

```
2018-01-03 16:14:10,613 ERROR [http-bio-8080-exec-13] o.r.C.Server [Slf4jLogger.java:230] An exception occurred writing the response entity
java.io.IOException: edu.illinois.library.cantaloupe.processor.ProcessorException: Unable to read the image header. (command output:
===========================================
The extension of this file is incorrect.
FOUND 5f1a. SHOULD BE .jp2
===========================================
imagetobmp: Unsupported precision: 2
[ERROR] Outfile /usr/local/fedora/tomcat/temp/cantaloupe-OpenJpegProcessor-2de6d695-0da1-42f8-86c7-1ad64d2acab1.bmp not generated
)
        at edu.illinois.library.cantaloupe.resource.ImageRepresentation.write(ImageRepresentation.java:125)
        at org.restlet.engine.adapter.ServerCall.writeResponseBody(ServerCall.java:513)
        at org.restlet.engine.adapter.ServerCall.sendResponse(ServerCall.java:457)
        at org.restlet.ext.servlet.internal.ServletCall.sendResponse(ServletCall.java:421)
        at org.restlet.engine.adapter.ServerAdapter.commit(ServerAdapter.java:191)
        at org.restlet.engine.adapter.HttpServerHelper.handle(HttpServerHelper.java:144)
        at org.restlet.ext.servlet.ServerServlet.service(ServerServlet.java:1117)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:727)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:303)
[...]

PR #202, opened 2018-02-20
Use separate symlinks per instance of OpenJpegProcessor.

adam-vessey:
Addressing part of #190, more specifically: Prevents the OpenJpegProcessor from breaking due to the the deletion of the intermediate symlink when attempting to process an unprocessable image by creating separate symlinks for each instance(/request).


adolski:
Works like magic. Thanks, @adam-vessey, for all of your work on this issue.

PR #217, opened 2018-05-22
Respect X-Forwarded-* headers

jweisman:
Respect any `X-Forwarded-*` headers that are provided. Currently headers
are respected only if `X-Forwarded-Host` header is also provided.

AWS [Elastic Load Balancing](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html), for example, sends `Proto` and `Port` but not `Host`.

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/resource/AbstractResource.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/resource/AbstractResource.java
@@ -98,7 +98,7 @@ public abstract class AbstractResource extends ServerResource {
         if (".".equals(appRootRelativePath)) { // when the paths are the same
             appRootRelativePath = "";
         }
-        Reference newRef = requestRef;
+        Reference newRef = new Reference(requestRef);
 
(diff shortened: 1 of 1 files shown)


src/main/java/edu/illinois/library/cantaloupe/resource/AbstractResource.java
```diff
@@ -98,7 +98,7 @@ protected static Reference getPublicReference(Reference requestRootRef,
         if (".".equals(appRootRelativePath)) { // when the paths are the same
             appRootRelativePath = "";
         }
-        Reference newRef;
+        Reference newRef = requestRef;
```
adolski: Could you do `Reference newRef = new Reference(requestRef);` here to avoid modifying `requestRef`?

jweisman:
Thanks @adolski. I committed the additional changes.

adolski:
Excellent. Thanks, @jweisman.

PR #260, opened 2019-01-10
intermediate response listener to handle exceeded buffer capacity

wwelling:


src/main/java/edu/illinois/library/cantaloupe/source/HttpSource.java
```diff
@@ -410,17 +411,21 @@ private void fetchRangedGETResponse() throws IOException {
...
+
+                Response response = listener.get(getRequestTimeout(), TimeUnit.SECONDS);
+
                 rangedGETResponseStatus = response.getStatus();
                 rangedGETResponseHeaders = response.getHeaders();
-                rangedGETResponseEntity = response.getContent();
-            } catch (ExecutionException e) {
-                throw new AccessDeniedException(info.getURI().toString());
-            } catch (InterruptedException | TimeoutException e) {
+                rangedGETResponseEntity = IOUtils.toByteArray(listener.getInputStream());
```
adolski: This InputStream needs to be closed I think

wwelling: Done using try-with-resources.

Issue #267, opened 2019-02-06
standalone war cannot start using the embedded servlet container, throws java.lang.NoClassDefFoundError for multiple classes


hardyoyo:
Currently, after compiling following the directions in the readme, the develp branch produces a standalone war file which throws java.lang.NoClassDefFoundError for multiple classes. One class is StringUtils.class, which is required be in the standalone war in order for it to function. It needs to be included much the same as [other classes are already copied to the root of the war file layout](https://github.com/medusa-project/cantaloupe/blob/develop/pom.xml#L449-L460). In so doing, more classes may be uncovered that need similar treatment, or to be refactored out. This ticket should track that work.

hardyoyo:
I revised the title of this issue to reflect the real issue I'm trying to solve, and so I can justify more work on my pull request: https://github.com/medusa-project/cantaloupe/pull/268

PR #268, opened 2019-02-06
fix standalone war so that it can again launch as documented

hardyoyo:
This pull request will attempt to fix the issue detailed here: https://github.com/medusa-project/cantaloupe/issues/267

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -504,6 +504,7 @@
                     <include name="edu/illinois/library/cantaloupe/util/FilesystemWatcher.class" />
                     <include name="edu/illinois/library/cantaloupe/util/FilesystemWatcher$Callback.class" />
                     <include name="edu/illinois/library/cantaloupe/util/SystemUtils.class" />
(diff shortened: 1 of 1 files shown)


adolski:
Thanks for submitting this, @hardyoyo. I haven't looked at it yet, but I think that adding the logger to the standalone build is not going to work, and it would be better to remove the dependency on the logger from StringUtils.

hardyoyo:
Ah, OK, I'll leave you to it, then. We're in the middle of troubleshooting why we cannot get the standalone war to run after compiling it from the HEAD of the develop branch, and this looked like the best approach. I can drop the logger commit from this, if that's the only objection?

adolski:
Sorry about that. Yes, the logger commit should be removed and any references to the logger from StringUtils would also have to be removed (the two `org.slf4j` imports, the `private static final Logger` assignment, and the usage within `md5()`).

hardyoyo:
I may not be free to work on #267 or #268 any time soon, so I'd not be offended if anyone else wanted to run off with this work from where I've left it.

Issue #303, opened 2019-08-02
JPEG2000KakaduImageReader causes JVM to crash For Certain JP2 images


bbarberBPL:
The JPEG2000KakaduImageReader causes the JVM and produces a core dump log file.  This does not occur with all images.  The following error occurs in the logs right before the JVM crashes

The errors are similar to [Issue 300](https://github.com/cantaloupe-project/cantaloupe/issues/300)

`ERROR [qtp920011586-1304] e.i.l.c.p.c.JPEG2000KakaduImageReader [JPEG2000KakaduImageReader.java:157] Kakadu Core Error:
You may not apply restrictions to the resolution or number of image components
after the first tile access, unless the codestream object is set up to be
persistent.`

Here is the core dump file.
[hs_err_pid2152.log](https://github.com/cantaloupe-project/cantaloupe/files/3462341/hs_err_pid2152.log)

Also here are the the test images as well as some log snippets for testing.
[cantaloupe-413-badimg2.zip](https://github.com/cantaloupe-project/cantaloupe/files/3462350/cantaloupe-413-badimg2.zip)

adolski:
Hi @bbarberBPL,

I can't reproduce this. Are you sure that the images in the zip file are the ones associated with the crashes?

For example, one of the errors has thread ID `qtp920011586-58.` Did this thread handle other requests after the one for `/commonwealth:cn69mp97w/full/,500/0/default.jpg`?

bbarberBPL:
Actually I have a more recent one for you
the url it was hitting was `/full/pct:30/0/default.jpg`

[bad_img.jp2.zip](https://github.com/cantaloupe-project/cantaloupe/files/3462760/bad_img.jp2.zip)

adolski:
That one works for me too.

bbarberBPL:
I'm not sure what happened with those initial ones. It could have been how I was grepping the logs. I'll try and find more before the end of the day here. Otherwise I can give them to you Monday

adolski:
Okay, thanks. I'll be out next week so take your time.

bbarberBPL:
@adolski  So we are having issues reproducing the issues on our end as well. However I think i might have narrowed down the cause. 

In the core dump file the stack shows the following 

```Stack: [0x00007fd47626e000,0x00007fd47636f000],  sp=0x00007fd47636c4c0,  free space=1017k
[...]

PR #308, opened 2019-08-20
Fixed KakaduDemoProcessor Memory Issue By Changing files to .bmp

bbarberBPL:
Hello @adolski. Since we were having those issues with the KakaduNativeProcessor, outlined in issue [#303](https://github.com/cantaloupe-project/cantaloupe/issues/303), with the native processor in release/4.1  I decided to roll the version back to release/4.0 and use the KakaduDemoProcessor. I rolled the version back to this version because I know you are in the process of deprecating the demo processor in version 4.1 plus It wasn't working correctly anyways(Probably due to the deprecation). 

Once I switched and got everything set up I found another bug right away when using the demo processor. The error being thrown was 

`java.io.IOException: I/O error reading header! (command output: terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
)`
It seemed to be coming from Kakadu itself. I verified this by running this in the command line.

` ./kdu_expand -quiet -resilient -no_alpha -i /var/cache/cantaloupe/source/6a/11/c8/6a11c8a353692e0c72a388449316b3d0 -o /tmp/cantaloupe/KakaduDemoProcessor-37ace037-ffa3-4740-a5c6-4d8adaf5afc4.tif`

This threw the same error code. After combing through the Kakadu docs and online. I found that the error is thrown do to not enough memory being allocated to read/ write the file. Since tifs are large files anyways, I discovered the next best option was to convert it to a bmp. After running the tests and producing a maven build I found that everything works perfectly now. No crashes, errors, and good performance. Here are the changes to facilitate this on release/4.0

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/processor/KakaduDemoProcessor.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/processor/KakaduDemoProcessor.java
@@ -136,9 +136,9 @@ class KakaduDemoProcessor extends AbstractJava2DProcessor implements FileProcess
      */
     private static void createStdoutSymlink() throws IOException {
         Path tempDir = Application.getTempPath();
-
+        ///changed this to .bmp as .tif was throwing std:bad_alloc errors
         final String name = KakaduDemoProcessor.class.getSimpleName() + "-" +
-                UUID.randomUUID() + ".tif";
+                UUID.randomUUID() + ".bmp";
         final Path link = tempDir.resolve(name);
         final Path devStdout = Paths.get("/dev/stdout");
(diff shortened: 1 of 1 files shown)


adolski:
Hi @bbarberBPL,

Efficiency-wise, BMP would be preferable to use in KDP because it can be streamed from the process whereas TIFF has to be fully written into a memory buffer first. (I suspect the `std::bad_alloc` is happening somewhere around there.)

But, kdu_expand doesn't color-adjust the pixels in its BMP output, nor does it embed an ICC profile. That's why TIFF is chosen.

PR #316, opened 2019-09-05
Added logstash configuration

cmhdave:
Just added the bare minimum for error and application logs.  Due to how access logs are created it takes a bit more to add logback-access to the project so I left that out for now.  I also opted not to enable `<includeCallerData>true</includeCallerData>` as the Logstash Logback Encoder documentation says that is a costly option to have turned on in a production environment.  It's possible to have a "debug" version of this at some time as well.

This implementation assumes a logstash shipper running on the server alongside Cantaloupe which is outside the scope of this PR.  Generally just making a JSON log file that's ready to be sent off to an ELK stack.  Obviously more configuration can be added to make this more robust but for now I felt this was the best bang with the fewest changes.

### DIFF
--- a/cantaloupe.properties.sample
+++ b/cantaloupe.properties.sample
@@ -668,9 +668,11 @@ log.application.ConsoleAppender.enabled = true
 # N.B.: Don't enable FileAppender and RollingFileAppender simultaneously!
 log.application.FileAppender.enabled = false
 log.application.FileAppender.pathname = /path/to/logs/application.log
+log.application.FileAppender.logstash.enabled = false
 
 log.application.RollingFileAppender.enabled = false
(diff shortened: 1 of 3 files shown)


adolski:
It's going to be a while before I can take a look, but thanks, @cmhdave .

adolski:
This looks good. Would there be any benefit to adding logstash output to the application log ConsoleAppender?

cmhdave:
> Would there be any benefit to adding logstash output to the application log ConsoleAppender?

I don't think so.  It's straight JSON and not something you would want to see in the console.

adolski:
What if the app is running in (say) a Docker container where the logging would ordinarily go to stdout/stderr?

PR #324, opened 2019-10-04
Add scaled option for image overlays

### LINKED ISSUE #322: Enhancement to have a watermark overlay image scaled to fit the entire base image
We have a need to have our watermark overlay image scaled to fit the entire image.  The reason being, images that are hosted come in all different sizes so it's not practical to have one-size watermark image.  Using `repeat` is an option but if you have a HUGE image in your collection and your watermark image is not very big, you get way too many watermarks and it looks kind of comical.

An option we'd like to contribute is a `scaled` option for watermark where the overlay image is scaled to fit inside the image and centered either horizontally or vertically based on the layout of the image compared to the layout of the overlay.

cmhdave:
This pull request resolves #322


cmhdave:
I found an issue with this code, don't merge as I need to fix it.

adolski:
Thanks @cmhdave! Great work. Sorry for the delay.

Issue #331, opened 2019-11-19
Update dependencies flagged in OWASP Dependency Check


cmhdave:
Part of our analysis for use of Cantaloupe as a IIIF server was to include the OWASP Dependency Check maven plugin into the pom.xml and run a report.  Having done that, several of the dependencies in the application were flagged as having outstanding CVEs that are resolved in newer versions.  We are currently running a special build with updated dependencies to see if everything still works after upgrading the dependencies but it might be nice to incorporate these changes back to the mainline project.

PR #332, opened 2019-11-19
Update dependencies to addres issue #331

cmhdave:
I've updated the dependencies as far as I can and still get a clean build from Travis CI.  In particular updating Jetty to the latest version does cause a build failure which is why it's only upgraded to 9.4.20


adolski:
Thanks @cmhdave!

PR #335, opened 2019-12-07
Configuration should return null if key is present but value is empty #333

### LINKED ISSUE #333: Since 4.1.0, fallback AWS credential providers below the Cantaloupe config provider cannot be used
**TL;DR:** Cantaloupe 4.1+ uses empty AWS credentials if `EnvironmentVariableCredentialsProvider` and `SystemPropertiesCredentialsProvider` fail and the config fallback is not specified. This is because the config fallback accepts empty credentials. This means that `ProfileCredentialsProvider` and `EC2ContainerCredentialsProviderWrapper` are never checked.

**The Problem**
I'm trying to deploy Cantaloupe `4.1.4` on ECS as with a task role, but I cannot set up authentication in the usual way since the `EC2ContainerCredentialsProviderWrapper` is never checked for credentials.

I believe that this is because since 4.1.0 (specifically, [this commit](https://github.com/cantaloupe-project/cantaloupe/commit/8b6ff77911812b434c386f7b50afbee0f98a2b2d#diff-4af9f42ceb870e2e6f21cdf208bd3484)), empty credentials have not been guarded against as [they used to be](https://github.com/cantaloupe-project/cantaloupe/blob/v4.0/src/main/java/edu/illinois/library/cantaloupe/util/AWSClientBuilder.java#L131). Since I want to use `EC2ContainerCredentialsProviderWrapper`, I leave the Cantaloupe config blank but it is used anyway. 🍈

MorrisonCole:
Tl;DR: Fixes #333 

There are a few different ways to fix #333 but I feel like this is the root cause. Since Cantaloupe ships an `example.properties` file with keys having no value, it seemed to me that the intended configuration API was broken.

If a key exists in configuration but has no value, it now returns null, whereas before the behaviour depended on the property type (e.g., String would return `""`). This brings the behaviour in line with environment properties, etc. as is standard.

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/util/StringUtils.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/util/StringUtils.java
@@ -3,6 +3,7 @@ package edu.illinois.library.cantaloupe.util;
 import edu.illinois.library.cantaloupe.config.Configuration;
 import edu.illinois.library.cantaloupe.config.Key;
 
+import javax.annotation.Nullable;
 import javax.xml.bind.DatatypeConverter;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
@@ -148,7 +149,11 @@ public final class StringUtils {
      * @return Boolean value of the given string.
      * @throws NumberFormatException if the string has an unrecognized format.
      */
-    public static boolean toBoolean(String str) {
+    public static boolean toBoolean(@Nullable String str) {
+        if (str == null) {
+            throw new NumberFormatException();
+        }
+
         switch (str) {
             case "1":
             case "true":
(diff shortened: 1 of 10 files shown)


src/main/java/edu/illinois/library/cantaloupe/util/StringUtils.java
```diff
@@ -148,7 +149,11 @@ public static String stripStart(String str, String toStrip) {
      * @return Boolean value of the given string.
      * @throws NumberFormatException if the string has an unrecognized format.
      */
-    public static boolean toBoolean(String str) {
+    public static boolean toBoolean(@Nullable String str) {
+        if (str == null) {
+            throw new NumberFormatException();
+        }
```
adolski: Could you add a test for this?

adolski:
Hi @MorrisonCole , thanks for submitting this. I made only one minor comment in the review. The other thing is that if you want this to go into the 4.1.x series, the PR should target the `release/4.1` branch instead of `develop`. Otherwise, great work!

MorrisonCole:
> Hi @MorrisonCole , thanks for submitting this. I made only one minor comment in the review. The other thing is that if you want this to go into the 4.1.x series, the PR should target the `release/4.1` branch instead of `develop`. Otherwise, great work!

Both good points! I'll rebase this and make a new PR for 4.1.x, then add the test you suggested 👍

Issue #340, opened 2019-12-19
Enable Jetty JMX for system monitoring and management


cmhdave:
In our environment we do quite a bit of our application monitoring and management via JMX beans.  It would be nice to include the out of the box Jetty JMX features into Cantaloupe so it too can be maintained and monitored via JMX

adolski:
Hi @cmhdave,

I'm OK with doing this but I'd want to make it clear that JMX is a Jetty feature not a Cantaloupe feature. It's possible (though not planned) that future versions of Cantaloupe won't use Jetty, and might use something else that doesn't support JMX. It would also be nice to be able to enable/disable JMX in the configuration with a default of disabled.

cmhdave:
Closing this issue as this is possible to enable through jdk command line options

cmhdave:
I've reopened this with the understanding that we don't need to make this a configurable option.  It will ONLY add Jetty JMX settings if the cantaloupe WAR was started with the JMX system property to allow for remote debugging ("com.sun.management.jmxremote.port").  This simplifies the implementation considerably.

PR #341, opened 2019-12-19
Add JMX beans for issue #340

cmhdave:


cmhdave:
As I think about this PR, it might make more sense to have this as a configurable option to turn it on.  I will modify the PR to include that and re-open the PR when it is done.

PR #354, opened 2020-02-28
Add Jetty JMX mbeans if cantaloupe started with JMX port

### LINKED ISSUE #340: Enable Jetty JMX for system monitoring and management

cmhdave:
This PR eliminates the need to add code to enable or disable JMX via configuration.  It simply just adds the jetty JMX features to the MBeanServer if Cantaloupe was started with a remote monitoring port (see "Remote Monitoring and Management" at https://docs.oracle.com/javase/7/docs/technotes/guides/management/agent.html)

fixes #340


src/main/java/edu/illinois/library/cantaloupe/ApplicationServer.java
```diff
@@ -281,6 +284,15 @@ public void start() throws Exception {
                 connector.setIdleTimeout(IDLE_TIMEOUT);
                 server.addConnector(connector);
             }
+
+            // If the Cantaloupe server is started with jmxremote, add the Jetty
+            // jmx extensions to the MBeanServer
+            if (System.getProperties().containsKey(REMOTE_JMX_PORT_PARAMETER) &&
+                    !"".equals(System.getProperty(REMOTE_JMX_PORT_PARAMETER))) {
```
adolski: Can you explain this condition to me? Does this mean that the server would have to be started with `-Dcom.sun.management.jmxremote.port` set to an empty value? I see that there are some other `com.sun.management.jmxremote.*` keys available, are those not needed or does MBeanContainer read them itself?

adolski: Sorry, I missed the "!". 🤦‍♂

Issue #355, opened 2020-03-03
JPEG2000MetadataReader hanging with KakaduDemoProcessor


cbeer:
With  a new install of Cantaloupe 4.1 configured to use `KakaduDemoProcessor` for JP2 (because we can't use `KakaduNativeProcessor` until #351 is fixed... it seems like we're seeing the same thing with `OpenJpegProcessor` too), we're seeing image and info requests get  hung up in [`readData()`](https://github.com/cantaloupe-project/cantaloupe/blob/v4.1.5/src/main/java/edu/illinois/library/cantaloupe/processor/codec/JPEG2000MetadataReader.java#L257-L260):

```
        while (readBox() != -1) {
            // Read boxes.
            isReadAttempted = true;
        }
```

We're seeing this with an image that has a UUID associated with it; other test images without the UUID seem to work fine. It looks like readUUIDBox may read only the first 8k of the box instead of the whole data stream.

PR #361, opened 2020-04-11
Use readFully instead of read for all JPEG2000MetadataReader operations

cbeer:
It turns out the issue reported  in #355 is more pervasive than just overly large UUID boxes and some unlucky JPEG2000 files exhibit the same problem when the buffer runs out.


adolski:
Sorry, I had to revert this merge because it broke the JPEG2000MetadataReader tests.

PR #383, opened 2020-06-26
Add support for microscopy images

stefano-malacrino:
This pull request adds support for microscopy images using the Bio-formats library.
The new classes `MicroscopyProcessor` and `MicroscopyImageReader` are implemented and the new `Format.Type` `MICROSCOPY` is defined.

### Supported formats
Bio-formats supports several image formats (see the list [here](https://docs.openmicroscopy.org/bio-formats/6.5.0/supported-formats.html)).
For now a `Format`  has been defined for the following pyramidal image formats

- Ventana BIF
- Zeiss CZI
- Bitplane Imaris
- Keller Lab Block
- Hamamatsu ndpi
- PerkinElmer Vectra QPTIFF
- Leica SCN
- Aperio SVS TIFF
- cellSens VSI

OME-TIFF images are also supported, but only using `ManualSelectionStrategy`. Beacuse they are TIFF files there is no way to detect them with `AutomaticSelectionStrategy`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -182,6 +182,13 @@
             <groupId>it.geosolutions.imageio-ext</groupId>
             <artifactId>imageio-ext-tiff</artifactId>
             <version>1.3.2</version>
+            <exclusions>
+                <!-- jai_imageio jar is already included in formats-gpl -->
+                <exclusion>
+                    <groupId>javax.media</groupId>
+                    <artifactId>jai_imageio</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>javax.servlet</groupId>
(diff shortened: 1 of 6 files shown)


adolski:
Thanks for submitting this, @stefano-malacrino, and sorry for the delay.

I've been thinking about what to do with this. The problem is that I'm not familiar with any of these formats, and don't have the resources to support them, or maintain this code. There is also a need for tests and documentation, but I can't help much with either of those for the same reasons.

I've been thinking about a way to improve the modularity of format support, which I have outlined in #388. This would involve replacing most of the Processors with format-specific [Image I/O plugins](https://docs.oracle.com/javase/8/docs/technotes/guides/imageio/spec/imageio_guideTOC.fm.html). It's going to take a lot of work to replace the existing Processors, but new plugins could be added even now with only minor changes. And I think that that is more ideal for a case like this because it would take the burden of format support off of Cantaloupe and move it to another party that is better equipped to handle it.

stefano-malacrino:
Hi @adolski,

Thank you for your comment. So you suggest that I implement a separate Image I/O plugin around Bio-formats instead of using a Processor?

adolski:
Hi @stefano-malacrino, yes, I think that is the way to go.

PR #391, opened 2020-08-06
Bugfix for error when delegate requests nil overlay

ccare:
Hello, 

A new user here - and very much enjoying Cantaloupe so far. Thank you!

I'm not sure how much you welcome patches, but I've been using the latest 4.1 release and had problems when attempting to control an overlay with a delegate method. The docs indicate that to disable an overlay, the delegate method should return `nil`, however when I tried this, I got a nasty stack trace from within `OperationList` about a null operation.

The fix is to guard against null, but I wasn't sure where in the call stack to fix it, and not being familiar with the code base, I'm not sure what your local styles are. But I noticed two things:

1) that the `OperationList.add()` method already had a non-null check, so I thought adding them to `addAfter` and `addBefore` would be useful. This PR has a commit to do that.

2) it felt like the specific case should be guarded too (although I wasn't sure about this)  and I've added another commit for that.

Hope it's helpful.

Thanks,

Charles


adolski:
Hi @ccare , thank you for contributing this!

Issue #418, opened 2020-12-08
scale_constraint off-by-one problem


andrewmccarthy:
Cantaloupe 4.1.7:

I'm using `delegates.rb` as described in the docs to set the `scale_constraint` limit to 1/2.

The original image is 4905×6788, and the constrained info.json gives the largest image as 2453×3394. However, if I try to download the full area in that size I get error "403 Access denied for the requested scale." I can, though, download 2452×3394.

It looks to me like one part of Cantaloupe is rounding 2452.5 up, while the other rounds down.

Thanks!

adolski:
Hi @andrewmccarthy, thanks for reporting this. It is now fixed on the `release/4.1` branch, for inclusion in 4.1.8.

PR #425, opened 2021-01-02
Fix scale constraints checking

kaij:
@adolski I think commit 975c3171976d3c81fd865268fe16510d20032c19 in #418 is breaking scale constraints in the allowed range in the develop branch. As soon as the effective scale s is smaller than the scale constraint minus delta, the exception is triggered. I attached a test which should run fine (s = 0.1), but throws IllegalScale Access denied.

Added two tests and tried to fix this in the expression. Please have a look at it - I'm not completely sure it handles all cases (although all tests run through now). E.g. I don't know if it's possible to specify negative scale constraints in the url (then it would need the Math.abs)

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/operation/OperationList.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/operation/OperationList.java
@@ -785,7 +785,6 @@ public final class OperationList implements Iterable<Operation> {
             }
         }
 
-        System.out.println("----");
         // Ensure that the resulting pixel area is positive.
         if (resultingSize.isEmpty()) {
             throw new ValidationException("Resulting pixel area is empty.");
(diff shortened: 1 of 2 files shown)


src/main/java/edu/illinois/library/cantaloupe/operation/OperationList.java
```diff
@@ -776,14 +776,16 @@ public void validate(Dimension fullSize,
...
             if (Arrays.stream(scales)
-                    .filter(s -> Math.abs(s - scaleConstraint.getRational().doubleValue()) > delta)
+                    .filter(s -> s - delta > scaleConstraint.getRational().doubleValue())
                     .findAny()
                     .isPresent()) {
                 throw new IllegalScaleException();
             }
         }
 
+        System.out.println("----");
```
adolski: Please remove

adolski:
Thanks for taking care of this @kaij!

PR #429, opened 2021-01-14
grok: upgrade to version v7.6.5

boxerab:
7.6.5 fixes a number of security bugs and crashes relative to 7.6.1

### DIFF
--- a/docker/Linux-GraalVM20/Dockerfile
+++ b/docker/Linux-GraalVM20/Dockerfile
@@ -12,6 +12,14 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 		maven \
 		wget \
 		libopenjp2-tools \
+                liblcms2-dev \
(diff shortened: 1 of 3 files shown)


boxerab:
hi @adolski , is there any way of getting Perl 5.32 installed on the graalvm VM ?

adolski:
I don't understand why it's complaining about perl in the GraalVM image, but not in any of the other images that are based on the same `ubuntu:latest` image. They all appear to have perl 5.30 installed.

boxerab:
Yes, that is odd. I looked into my debian packages, and they are built against Perl 5.32 .so binaries, but if this is a problem, it should appear in all VMs as you mentioned.

boxerab:
@adolski I re-built on a system with perl 5.30. Now ready to merge.

boxerab:
Thanks!

PR #434, opened 2021-02-08
Add PDFBox configuration options

### LINKED ISSUE #432: PDFBox loads entire PDF into memory to render single page
We have found that some extremely large PDFs (think over 1GB) can very quickly exhaust all server memory resources if multiple requests come in for different pages of the same PDF.  For example if you create a presentation manifest with a separate entry for each page of the PDF, in something like Mirador you would render multiple pages at once and quickly gobble up many multiple GB as it loads the entire PDF into memory for EACH page.

PDFBox as implemented in Cantaloupe's PdfBoxProcessor calls `PDDocument.load()` with a single parameter, either the file or an input stream.  PDFBox has the ability to also tell it how to use memory:
```
PDDocument load(File file, MemoryUsageSetting memUsageSetting)
```
or
```
PDDocument load(InputStream input, MemoryUsageSetting memUsageSetting)
```
`MemoryUsageSetting` allows you to set these properties to better control memory consumption
```
    private final boolean useMainMemory;
    private final boolean useTempFile;
    private final long maxMainMemoryBytes;
    private final long maxStorageBytes;
    private File tempDir;
```
I feel it has become critical for us to be able to control these settings via Cantaloupe configuration.  Something like this 

```
MemoryUsageSetting settings = MemoryUsageSetting.setupMixed(maxMemoryBytes).setTempDir(pdfTempDir);
doc = PDDocument.load(sourceFile.toFile(), settings);
```

I could take a swing at a PR. Or if you want to address it I'll stand back.  I just don't want to duplicate effort.

cmhdave:
Resolves #432

This introduces settings to control how PdfBoxProcessor loads PDFs. Currently the entire PDF is loaded into memory for each page.  If you have a very large PDF this will quickly lead to OutOfMemory issues.  PDFBox has the ability to share file loading across memory and a Scratch file on disk.


cmhdave:
I'm not entirely sure why this is failing.  Seems to fail on an S3 test which shouldn't be affected by the code changes I made.

adolski:
That's just a buggy test, don't worry about it. I will try to look at this soon.

src/main/java/edu/illinois/library/cantaloupe/processor/PdfBoxProcessor.java
```diff
@@ -187,6 +186,25 @@ public void put(COSObject indirect, PDXObject xobject) {
         }
     }
 
+    private MemoryUsageSetting getMemoryUsageSetting() {
+        final Configuration config = Configuration.getInstance();
+        final boolean scratchFileEnabled =
+                config.getBoolean(Key.PROCESSOR_PDF_SCRATCH_FILE_ENABLED, false);
+        final long maxMainMemoryBytes =
+                config.getLong(Key.PROCESSOR_PDF_MAX_MEMORY_BYTES, -1);
```
adolski: For keys with byte-size values, it's better to use `Configuration.getLongBytes()` which enables users to enter a value with K/KB, M/MB, etc. suffix

CHANGES.md
```diff
@@ -58,7 +58,12 @@
 * FfmpegProcessor, KakaduNativeProcessor, OpenJpegProcessor, and
   PdfBoxProcessor use libjpeg-turbo to write JPEGs if it is available.
 * Added the `scaled` position for image overlays.
-* Updated Apache PDFBox to version 2.0.21.
+* Updated Apache PDFBox to version 2.0.22.
+* PdfBoxProcessor properties added to better control memory
```
adolski, line 62: You can thank yourself here if you want :smile:

src/main/java/edu/illinois/library/cantaloupe/processor/PdfBoxProcessor.java
```diff
@@ -187,6 +186,25 @@ public void put(COSObject indirect, PDXObject xobject) {
...
+        final long maxStorageBytes =
+                config.getLong(Key.PROCESSOR_PDF_MAX_STORAGE_BYTES, -1);
+        final String scratchFileLocation =
+                config.getString(Key.PROCESSOR_PDF_SCRATCH_LOCATION);
+        if (scratchFileEnabled) {
+            File filePath = new File(scratchFileLocation);
+            return MemoryUsageSetting.setupMixed(maxMainMemoryBytes,
+                    maxStorageBytes).setTempDir(filePath);
+        } else {
+            return MemoryUsageSetting.setupMainMemoryOnly(maxMainMemoryBytes);
```
adolski: I noticed some unexpected behavior here. When I set `processor.pdf.scratch_file_enabled = false` and `processor.pdf.max_memory_bytes` to a small value, it may result in an incompletely rendered PDF. For example, I have a PDF in which all the text renders but the main image ends up blank. So I think it would be best not to constrain the in-memory size when scratch files are disabled.

CHANGES.md
```diff
@@ -58,7 +58,12 @@
...
+    * `processor.pdf.scratch_file_enabled` allows use of a scratch file
+    * `processor.pdf.max_memory_bytes` allows control over how much memory a PDF thread consumes (-1 means unlimited)
+    * `processor.pdf.max_storage_bytes` allows control over how much total bytes are used to render a PDF image (memory + temp file)
+    * `processor.pdf.scratch_location` allows customization of path to PDF scratch file
```
adolski: Could `temp_path` be used in place of this?

CHANGES.md
```diff
@@ -58,7 +58,12 @@
 * FfmpegProcessor, KakaduNativeProcessor, OpenJpegProcessor, and
   PdfBoxProcessor use libjpeg-turbo to write JPEGs if it is available.
 * Added the `scaled` position for image overlays.
-* Updated Apache PDFBox to version 2.0.21.
+* Updated Apache PDFBox to version 2.0.22.
+* PdfBoxProcessor properties added to better control memory
+    * `processor.pdf.scratch_file_enabled` allows use of a scratch file
+    * `processor.pdf.max_memory_bytes` allows control over how much memory a PDF thread consumes (-1 means unlimited)
+    * `processor.pdf.max_storage_bytes` allows control over how much total bytes are used to render a PDF image (memory + temp file)
```
adolski: In the interest of UX, is there any reason that this key couldn't be removed and max used by PDFBox set to unlimited?

adolski (changes requested):

src/main/java/edu/illinois/library/cantaloupe/processor/PdfBoxProcessor.java
cmhdave: Will fix.

CHANGES.md
cmhdave: You are absolutely correct.  I over-thought this because the scratch-file out-grew my system temp location but for some reason didn't think to just update the system temp location  I'll remove this setting and use `temp_path`

CHANGES.md
cmhdave: My only concern is not having some sort of failsafe on the amount of disk storage allocated to PDFBox... that being said, these are just temporary files that end up getting deleted as soon as the image is rendered and it wouldn't even be a failsafe in the case that many requests came in for a huge file (or requests for multiple pages from the same PDF via Mirador or Internet Archive book viewer).  In those cases there would still be temp files created for each image which could still go over the disk allocation.

Seems safe to remove and use -1 instead by default.

src/main/java/edu/illinois/library/cantaloupe/processor/PdfBoxProcessor.java
cmhdave: It seems as though for this you cannot have a `-1` value due to this line in `StringUtils`
```
final String numberStr = str.replaceAll("[^\\d.]", "");
```
It would need to be changed to this in order to allow `-1`
```
final String numberStr = str.replaceAll("[^-\\d.]", "");
```
I'm not sure what kind of side effect that may have (if any)

adolski: That should be fine

src/main/java/edu/illinois/library/cantaloupe/processor/PdfBoxProcessor.java
```diff
@@ -191,17 +191,15 @@ private MemoryUsageSetting getMemoryUsageSetting() {
         final boolean scratchFileEnabled =
                 config.getBoolean(Key.PROCESSOR_PDF_SCRATCH_FILE_ENABLED, false);
         final long maxMainMemoryBytes =
-                config.getLong(Key.PROCESSOR_PDF_MAX_MEMORY_BYTES, -1);
-        final long maxStorageBytes =
-                config.getLong(Key.PROCESSOR_PDF_MAX_STORAGE_BYTES, -1);
+                config.getLongBytes(Key.PROCESSOR_PDF_MAX_MEMORY_BYTES, -1);
         final String scratchFileLocation =
-                config.getString(Key.PROCESSOR_PDF_SCRATCH_LOCATION);
+                config.getString(Key.TEMP_PATHNAME);
```
adolski: Sorry, this is my fault for not thinking about it, but the way to get the canonical temp path is via `Application.getTempPath()` which will fall back to `java.io.tmpdir` if  `temp_path` is not set.

cmhdave: Nice clean way of handling that!  I'll update this now.

adolski:
Thanks @cmhdave !

Issue #233, opened 2018-08-29
HttpSource sends a GET Range request without checking if partial requests are supported


x1rskarw:
The proprietary HTTP server we wish to use as a source doesn't support partial requests and answers with full image data. 

HttpSourve.fetchRangedGETResponse() should check via a HEAD request first if a ranged requests makes sense at all, as described in: https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests

rprieto:
We are facing the same issue, thanks for fixing this! Do you have plans to release 4.0.3? This change doesn't seem to be in 4.0.2 that was released in October.

adolski:
Sorry, this was fixed on the `develop` branch which will become 4.1. I don't think it will be included in 4.0.x unless someone wants to send a PR.

PR #467, opened 2021-04-20
Add option to disable HEAD request

### LINKED ISSUE #465: Add ability to disable HEAD request on HttpSource
Hello. I have scenario where images can be served from different locations, e.g. filesystem, multiple S3 (MinIO), GCP.

We have special microservice which is responsible to return direct URL to a file by it's identifier. For images that stored on filesystem it returns URL to itself (for proxying file download), but for S3 and GCP it returns [PreSigned URLs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ShareObjectPreSignedURL.html), so other services can download file directly from S3 without transferring it through file microservice.

There is one problem with PreSigned URLs in Cantaloupe: when image is required, Cantaloupe maked `HEAD` request to URL which is returned by delegate to ensure that image is exists and available. But PreSigned URLs return 403 Forbidden error code because they allow only `GET` requests. Cantaloupe does not has any property to forbid outcoming HEAD requests, but such option can solve problem with PreSigned URLs in our scenario. For a `HEAD` request replacement, Cantaloupe can use `GET` request with small `Range` to fetch image info such as `Content-Type` and `Content-Length` if needed.

It can be a global option for HttpSource or additional key to return value of `httpsource_resource_info` delegate could be added, what do you think?

maksimkurb:
This PR adds option to disable HEAD request, so Cantaloupe can fetch images from S3 presigned URLs, which allows only GET requests and return 403 Forbidden for any HEAD request.

Fixes #465

### DIFF
--- a/cantaloupe.properties.sample
+++ b/cantaloupe.properties.sample
@@ -167,6 +167,10 @@ FilesystemSource.BasicLookupStrategy.path_suffix =
 # Trusts insecure certificates and cipher suites.
 HttpSource.allow_insecure = false
 
+# Use GET request with header Range = bytes=0-0 instead of
+# HEAD request. Useful if using Cantaloupe with presigned S3 URLs.
+HttpSource.disable_head_request = false
+
 # Request timeout in seconds.
 HttpSource.request_timeout =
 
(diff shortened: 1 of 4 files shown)


adolski:
Hi @maksimkurb, sorry for the delay.

HttpSource actually used to not use a HEAD request, but #233 changed this. Now it's clear that some servers will require a HEAD-first strategy whereas others (like yours) will require ranged GET.

I like your solution, but for more flexibility, I think I would like to make it work on a per-request basis which would mean adding:

1. A new config key, `HttpSource.BasicLookupStrategy.enable_head_requests` (not 100% sure on that name yet)
2. An `enable_head_request` key in the hash returned from `httpsource_resource_info()` (again not sure on name)

This will accommodate cases in which one source needs HEAD-first and another needs GET-first.

And it will be quite a bit more work than what you've done in #467, but I'm happy to do it, although it will take a while.

PR #499, opened 2021-06-07
Proxy feature

mightymax:
For a project that hosts a Nuxeo server behind a HTTP proxy, we need to be able to configure Cantaloupe to use this proxy in the HttpSource. The[ proxy is added to the OkHttpClient.Builder](https://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.html#proxy--) part. 

Three new config settings are introduced:
```properties
HttpSource.proxy.http.enabled = false
HttpSource.proxy.http.server = proxy.example.com
HttpSource.proxy.http.port = 8080
```
A RuntimeException is thrown when proxy is enabled and server is empty.


adolski:
Is it possible to add tests for this feature?

mightymax:
> Is it possible to add tests for this feature?

So sorry, I have looked at writing tests for this feature, but I have never before worked with Java. Adding Proxy support was relativaly easy for me, but writing the test is not I am afraid.

adolski:
Sorry for the delay, lots of other priorities lately. I've been experimenting with tests for this feature and I got them all passing. Could you resubmit this PR against the `develop` branch? (I prefer to keep the x.x.x releases bugfix-only.)

There are a few other minor changes I'd like to make as well, but I'm happy to take care of that based on what you've supplied.

PR #553, opened 2022-01-05
Feature/http proxy support

mightymax:
For a project that hosts a Nuxeo server behind a HTTP proxy, we need to be able to configure Cantaloupe to use this proxy in the HttpSource. The[ proxy is added to the OkHttpClient.Builder](https://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.html#proxy--) part. 

Three new config settings are introduced:
```properties
HttpSource.proxy.http.enabled = false
HttpSource.proxy.http.server = proxy.example.com
HttpSource.proxy.http.port = 8080
```
A RuntimeException is thrown when proxy is enabled and server is empty.

NB this is a duplicate for [this pull request](https://github.com/cantaloupe-project/cantaloupe/pull/499) by request of @adolski


adolski:
FYI, I renamed `HttpSource.proxy.http.server` to `HttpSource.proxy.http.host` and removed `HttpSource.proxy.http.enabled`, so the proxy will be enabled whenever a host is filled in.

PR #583, opened 2022-04-07
Allow delegate system to produce valid HTTP 401 info.json responses

### LINKED ISSUE #580: HTTP 401 info.json response is invalid
### The issue

As a follow up to #555: it looks like c9aac7550893a65a1fd5b35c7a796fab644bcb7c introduced a non-spec compliant change.

From https://iiif.io/api/auth/1.0/#interaction-with-access-controlled-resources (emphasis mine):
> These interactions rely on requests for Description Resources returning HTTP status codes 200, 302, and 401 in different circumstances. **In cases other than 302, the body of the response must be a valid Description Resource** because the client needs to see the Authentication service descriptions in order to follow the appropriate workflow.

[All versions of the Image API spec](https://iiif.io/api/image/3.0/#52-technical-properties) indicate that both `width` and `height`, for example, are required image information properties. Cantaloupe does not include those properties on the HTTP 401 responses, so the responses are not valid.

I have tested the change with Mirador 3, and it turns out that it throws an unrecoverable error, having made the assumption that the info.json in a 401 response has all the required properties.

### Side note

Further, while the Image API specs do not _require_ a particular content type for error responses, they [do _recommend_ "a human-readable description of the error in either plain text or html"](https://iiif.io/api/image/3.0/#73-error-conditions), so it should be considered that the addition of `status` and `message` to the 401 response introduces [the possibility of namespace collision](https://iiif.io/api/image/3.0/#57-extra-functionality) (for example, if someone came up with an Image API extension that uses those properties):
> [...] additional properties used in the image information, that are not defined in this specification should be mapped to RDF predicates using further context documents. These extensions should be added to the top level @context property (see Technical Properties). The JSON-LD 1.1 functionality of predicate specific context definitions, known as scoped contexts, must be used to minimize cross-extension collisions.

markmatney:
Closes #580. That issue attempts to explain why I think changing only the 401 responses for information requests is better than changing all 4xx responses.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## 5.0.6
 
+* IIIF information endpoints return JSON in HTTP 401 responses.
 * Fixed a bug whereby the values of the `operations` and `page_count` keys
   in the delegate context were not set.
 * TurboJpegProcessor is able to generate non-JPEG derivative images, which
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,6 @@
 
 ## 5.0.6
 
-* IIIF information endpoints always return JSON in HTTP 4xx responses.
 * Fixed a bug whereby the values of the `operations` and `page_count` keys
   in the delegate context were not set.
 * TurboJpegProcessor is able to generate non-JPEG derivative images, which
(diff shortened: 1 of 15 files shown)


cachemeoutside:
@adolski Any ETA on getting PR review and potentially released?

ksclarke:
Closing this PR because we've retargeted it to the `develop` branch (cf. #670)

PR #585, opened 2022-04-29
Fix HTTP 302 redirects via delegate for identifiers with slashes

### LINKED ISSUE #581: Delegate system does not allow HTTP 302 redirects for resources whose ID contains a slash
### The issue

If a resource's identifier contains a slash (for example, if it's an [ARK](https://en.wikipedia.org/wiki/Archival_Resource_Key)), then a 302 redirect triggered by a Cantaloupe delegate will result in a 404, even though the request would have succeed in absence of the redirect.

This error seems to occur regardless of the particular storage backend, although I have only attempted to reproduce with FilesystemSource and S3Source.

### To reproduce

1. Build the delegate here: https://github.com/markmatney/sample-java-delegate/tree/tiered-access-slash-in-id-mwe
2. Run a local Cantaloupe (I'm using 21057cf72071840e2a984450bf8845c6e6490ec9) with the delegate JAR on the classpath, and configure an image source that contains a test image inside at least one sub-directory (i.e. so that the image identifier contains at least one slash, e.g. `test/test.jpg`))
3. Request an info.json or image in your browser, using any of the Image API endpoints, until Cantaloupe attempts, via the delegate, a 302 redirect that will always result in an HTTP 404 response; in my minimal implementation, the redirect attempt will occur with 50-50 chance on each page load due to the "coin flip" in [this conditional statement](https://github.com/markmatney/sample-java-delegate/blob/00cdc2212e39c68558f7c6fa2c03caab137acd26/src/main/java/org/example/MyJavaDelegate.java#L29)
4. Note that when a redirect is _not_ attempted, the request succeeds; and when it _is_ attempted, the debug logs indicate that Cantaloupe _is_ able to find the resource before the redirect occurs, but not after, and the request ultimately results in a 404

markmatney:
Fixes #581.


cachemeoutside:
@adolski Any ETA on getting PR review and potentially released?

ksclarke:
Noting correspondence to UCLA docker-cantaloupe patch: issue-581-v5.0.5.patch

ksclarke (approved):

jcoyne:
I think it's great that you provided a couple of unit tests.  For us to understand the need better, can you provide a system test so that we can see when you have a certain system configuration, it gives a certain response.

glenrobson:
Discussed in the 27th November call that we would merge this and ask for Functional unit tests at a later time. UCLA have this patch in production as part of their Docker release.

PR #616, opened 2022-12-12
Avoid communicating with S3 unless necessary

garyttierney:
Previously the S3 source would eagerly connect to S3 retrieve information about the object being requested when creating a stream. This change defers retrieval of properties from S3 until needed by the S3StreamFactory, and allows cached files to be served directly from disk with no intermediate S3 lookups.

It's a small optimization, but can shave a dozen or so MS off of a cached image request.

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/source/S3ObjectInfoSupplier.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/source/S3ObjectInfoSupplier.java
@@ -1,0 +1,11 @@
+package edu.illinois.library.cantaloupe.source;
+
+import java.io.IOException;
+
+/**
+ * Deferred accessor to {@link S3ObjectInfo}.
+ */
+@FunctionalInterface
+public interface S3ObjectInfoSupplier {
+    S3ObjectInfo get() throws IOException;
+}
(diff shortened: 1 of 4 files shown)


DiegoPino:
@glenrobson this works and tested over current develop branch. But I would love to have your larger pull merged first and test with it and ideally give this a spin on a production server with real load to see if having a get Object Info from S3 fail last (when deferred) has any other consequences, if that is OK

DiegoPino:
@garyttierney hi. I rebased against `develop` and created a docker container using S3 as cache and source backend. All works well (testing what happens if something fails/goes wrong is hard to replicate so my testing is based on successful operations). It would be great if you could rebase this pull against https://github.com/cantaloupe-project/cantaloupe so we can let the CI tests run. Please let us know if you could tackle that in the few next days so I can approve and we move forward with merging. Thanks so much for your contribution!

garyttierney:
Yep no problem, rebased against latest changes in `develop`.

DiegoPino (approved):
Works and checks (finally!) are all passing

DiegoPino:
@garyttierney thanks for your contribution!

PR #617, opened 2022-12-12
Use ImageIO based OpenJPEG codec for JP2s

garyttierney:
More of a RFC/WIP. I've replaced the external process OpenJpegProcessor with a Java2dProcessor configured to use the [imageio-openjpeg](https://github.com/dbmdz/imageio-jnr) ImageIO plugin. This enables streaming for JP2 files and can avoid the bitmap disk copies, but isn't an ideal solution as mentioned in #563.

Some things I have to verify:
- Any uses of Metadata that was previously being constructed by Cantaloupe is compatible with the new ImageIO metadata object returned by imageio-openjpeg.
- Performance hasn't regressed for any cases


src/main/java/edu/illinois/library/cantaloupe/processor/AutomaticSelectionStrategy.java
```diff
@@ -11,8 +11,8 @@
 class AutomaticSelectionStrategy implements SelectionStrategy {
 
     private static final List<Class<? extends Processor>> JP2_CANDIDATES = List.of(
-            KakaduNativeProcessor.class,
```
DiegoPino, line 14: This here does not look right. KakaduNativeProcessor.class should not be removed, but instead OpenJpegProcessor.class if Java2d is the one handling it now

DiegoPino (changes requested):
Can't speak for the actual code yet, but there is an error in the automatic selection strategy commit of this pull

src/main/java/edu/illinois/library/cantaloupe/processor/AutomaticSelectionStrategy.java
garyttierney: Oh, good catch, thanks.

jcoyne:
@garyttierney what are your intentions here?  Are you planning on converting this to an actual pull request or can we close this draft now?

garyttierney:
> @garyttierney what are your intentions here? Are you planning on converting this to an actual pull request or can we close this draft now?

I think we can close it for now and I'll revisit if there's a desire again to get streaming working with OpenJPEG. With the foreign function & memory interfaces stabilized in Java 22 it's now probably worth exploring a Project Panama based binding that can avoid the large amount of buffer copies the current implementations do.

PR #629, opened 2023-04-13
Upgrade jruby to v9.3.5.0 to support M1 Mac

### LINKED ISSUE #628: Upgrade jruby to support M1 Mac
Current JRuby version relies on a JFFI that does not support M1 Mac arm64.
According to https://www.jruby.org/2022/01/19/jruby-9-3-3-0.html versions since 9.3.3 do support M1 Mac
> The big news this release is support for Apple’s M1 family of processors (Apple Silicon) via an update to our native FFI backend library.

The error msg I'm getting is
> could not get native definition for type `POINTER`, original error message follows: java.lang.UnsatisfiedLinkError: Unable to execute or load jffi binary stub from `/var/folders/hb/_6h0qb492dzg72h1n0gw7fnw0000gn/T/`. Set `TMPDIR` or Java property `java.io.tmpdir` to a read/write path that is not mounted "noexec".

janhoy:
Closes #628 

Not yet tested IRL but seems to build

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,7 +18,7 @@
         <aws-sdk.version>2.15.28</aws-sdk.version>
         <jackson.version>2.11.0</jackson.version>
         <jetty.version>11.0.5</jetty.version>
-        <jruby.version>9.2.17.0</jruby.version>
+        <jruby.version>9.3.5.0</jruby.version>
         <logback.version>1.2.8</logback.version>
         <surefire.version>3.0.0-M3</surefire.version>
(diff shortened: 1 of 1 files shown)


glenrobson:
I believe this is now done in 5.0.6: https://github.com/cantaloupe-project/cantaloupe/blob/release/5.0/pom.xml

janhoy:
Ah, I forgot to check. Thanks, will take it for a spin.

Issue #593, opened 2022-06-28
tiling problem for PDF on y axis


hrvoj3e:
Related to this problem
 - https://github.com/cantaloupe-project/cantaloupe/issues/482

https://github.com/cantaloupe-project/cantaloupe/issues/482#issuecomment-1103813724

> I tried sample.pdf with this commit https://github.com/cantaloupe-project/cantaloupe/commit/fff54250eda8e5399ae59845ca4ec4f299bb3439 and it mostly works but not for images with x=0 on y axis.
> 
> E.g.
> 0,1024,512,512: /iiif/2/https:%2F%2Fgithub.com%2Fcantaloupe-project%2Fcantaloupe%2Ffiles%2F6443024%2Fsample.pdf/0,1024,512,512/512,/0/default.jpg?cache=nocache
> 
> will return the same image as
> 
> 0,0,512,512: /iiif/2/https:%2F%2Fgithub.com%2Fcantaloupe-project%2Fcantaloupe%2Ffiles%2F6443024%2Fsample.pdf/0,0,512,512/512,/0/default.jpg?cache=nocache
> 
> @adolski could you check please?

Is there a fix for this?
Which commit should I target for this?

La0:
We also encounter this problem. X & Y coordinates are ignored when using PDF files in cantaloupe

DiegoPino:
See https://github.com/cantaloupe-project/cantaloupe/issues/482#issuecomment-1674917947

DiegoPino:
Hi @hrvoj3e testing went well. TurboJPEG enabled + proper Tiling with X and Y offsets (and 0 too of course). I have trace logs enabled and all virtual cropping looks good (enough).

If you want to build your own here is our branch https://github.com/esmero/cantaloupe/tree/ISSUE-593
I also pushed a new container to our own list of too many Cantaloupe containers:
`esmero/cantaloupe-s3:6.0.1-multiarch`, that is the repo/name and tag. Multiarch because it works on ARM64 and AMD/Intel64. has Grok, Kakadu (only works on x86, that is because Kakadu does not provide a public binary for arm64), ffmpeg, turbojpeg, etc.

This is how we use (for developing/testing) the container in a `docker-compose` mix. Swap the tag there and you need to provide your own cantaloupe properties files of course (the example mounts it from a folder, just change your path)
[...]

Issue #482, opened 2021-05-07
TurboJPEG has problems tiling images from PDFBox


cmhdave:
We are seeing issues with TurboJPEG and tiling PDFs.  It is really strange and arbitrary.  What I can say is that it does not happen when you remove TurboJPEG from the processing chain.  I am able to reproduce this every time with the [sample.pdf](https://github.com/cantaloupe-project/cantaloupe/files/6443024/sample.pdf) file that I've created and attached.  

What you need to do is start Cantaloupe 5.0 with TurboJPEG available in the library path.   We do this like so:
```
java -cp cantaloupe.jar -Djava.library.path=/usr/local/opt/jpeg-turbo/lib -Dcantaloupe.config=cantaloupe.properties edu.illinois.library.cantaloupe.StandaloneEntry
```
Then you open a tile like below (which happens to be a tile that was being requested from our manifest via Mirador)

http://localhost:8182/iiif/2/sample.pdf/512,512,512,512/512,/0/default.jpg

The resulting image starts at x=0, y=0 instead of x=512, y=512.

Now try the same URL with a PNG extension and you will get the correct tile.

http://localhost:8182/iiif/2/sample.pdf/512,512,512,512/512,/0/default.png

Here's where it gets strange.  If you load the PDF as a JPG with a slightly different size it works correctly (notice `513,` rather than `512,`

http://localhost:8182/iiif/2/sample.pdf/512,512,512,512/513,/0/default.jpg

Another strange thing is that I can debug this in IntelliJ into `TurboJPEGImageWriter` in `write(BufferedImage image, OutputStream os)` where when you call `tjc.setSourceImage(image, 0, 0, 0, 0)` the image IS THE CORRECT TILE.

![image](https://user-images.githubusercontent.com/32506/117480954-496b6c80-af30-11eb-9f52-c147fee08c69.png)

![image](https://user-images.githubusercontent.com/32506/117481094-7a4ba180-af30-11eb-9abf-ade170afa80a.png)

Yet this is what gets rendered in the browser.

![image](https://user-images.githubusercontent.com/32506/117481282-c4348780-af30-11eb-93c2-ab458937e1a9.png)

I have all very basic controls turned on and no caching enabled (to rule out an caching issue)

========

[sample.pdf](https://github.com/cantaloupe-project/cantaloupe/files/6443024/sample.pdf)

adolski:
@cmhdave Could you try this in 5.0.2?

[...]

PR #655, opened 2024-05-08
ISSUE-593: Fix Virtual Cropping detection condition to take both Y and X in account

DiegoPino:
# What?

See #593

Fixes a bug generated by a small typo. Without this fix, PDF tile generation will only result in a single axis being tiled.


ksclarke (approved):

glenrobson:
Just looking through the original #482 issue it looks like this was fixed in 5.0.2 but came back in 5.0.4. Although its a simple fix I think it would be useful to include a test to make sure this doesn't get re-introduced.

DiegoPino:
@glenrobson my understanding is it was not completely fixed by https://github.com/cantaloupe-project/cantaloupe/commit/fff54250eda8e5399ae59845ca4ec4f299bb3439 (typo in that fix) thus my pull. I can write a test but only by July 2024 (sorry), my June availability because of day-work and conferences is very limited.

glenrobson:
Glen to add to test.

PR #551, opened 2022-01-03
Fix buffer overflow when reading EXIF metadata

La0:
Closes #550

Issue #550, opened 2022-01-03
Buffer overflow when reading some EXIF metadata


La0:
I stumbled upon some images that trigger a server internal error when cantaloupe tries to read their EXIF metadata.

Here is the stacktrace

```
java.lang.IndexOutOfBoundsException: off < 0 || len < 0 || off+len > b.length || off+len < 0!
	at java.desktop/javax.imageio.stream.MemoryCacheImageInputStream.read(MemoryCacheImageInputStream.java:100)
	at edu.illinois.library.cantaloupe.image.exif.Reader.readBytes(Reader.java:159)
	at edu.illinois.library.cantaloupe.image.exif.Reader.read(Reader.java:145)
	at edu.illinois.library.cantaloupe.image.exif.Reader.read(Reader.java:97)
	at edu.illinois.library.cantaloupe.processor.TurboJpegProcessor.readMetadata(TurboJpegProcessor.java:275)
	at edu.illinois.library.cantaloupe.processor.TurboJpegProcessor.readInfo(TurboJpegProcessor.java:263)
	at edu.illinois.library.cantaloupe.cache.InfoService.readInfo(InfoService.java:220)
	at edu.illinois.library.cantaloupe.cache.InfoService.getOrReadInfo(InfoService.java:153)
	at edu.illinois.library.cantaloupe.cache.CacheFacade.getOrReadInfo(CacheFacade.java:67)
	at edu.illinois.library.cantaloupe.resource.AbstractRequestHandler.getOrReadInfo(AbstractRequestHandler.java:39)
	at edu.illinois.library.cantaloupe.resource.InformationRequestHandler.handle(InformationRequestHandler.java:287)
	at edu.illinois.library.cantaloupe.resource.iiif.v2.InformationResource.doGET(InformationResource.java:85)
	at edu.illinois.library.cantaloupe.resource.HandlerServlet.handle(HandlerServlet.java:97)
	at edu.illinois.library.cantaloupe.resource.HandlerServlet.doGet(HandlerServlet.java:35)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:687)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:791)
	at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:550)
[...]

PR #656, opened 2024-05-09
Merge in updated libraries from Cantaloup 5.0.6 and fix test failures

angelahuqing:
5.0.6 is merged into develop branch as well as replacing or removing tests/functions that aren't being used and causing failure.

PR #653, opened 2024-05-08
Merge in updated libraries from Cantaloup 5.0.6

glenrobson:

PR #659, opened 2024-06-06
Fix some unsafe reads.

adam-vessey:
If they failed to read the number of bytes, the return of `-1` would cause infinite loops, with `-1` always less than the minimum `offset` of `0`.

An alternative to #551 fixing #550, and more, touching on https://github.com/cantaloupe-project/cantaloupe/pull/551#discussion_r1427377512 

Did a bit of a `grep` around for similar broken patterns which should get a similar fix; though, there may be some others. In particular, suspicious of [this bit of `JPEG2000KakaduImageReader`](https://github.com/cantaloupe-project/cantaloupe/blob/21057cf72071840e2a984450bf8845c6e6490ec9/src/main/java/edu/illinois/library/cantaloupe/processor/codec/jpeg2000/JPEG2000KakaduImageReader.java#L103-L110), but not terribly sure on the expectations around it, and don't use Kakadu.

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/image/exif/Reader.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/image/exif/Reader.java
@@ -155,10 +155,7 @@ public final class Reader implements AutoCloseable {
 
     private byte[] readBytes(int length) throws IOException {
         byte[] data = new byte[length];
-        int n, offset = 0;
-        while ((n = inputStream.read(data, offset, data.length - offset)) < offset) {
-            offset += n;
-        }
+        inputStream.readFully(data);
         return data;
     }
 
--- a/src/main/java/edu/illinois/library/cantaloupe/processor/codec/gif/GIFMetadataReader.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/processor/codec/gif/GIFMetadataReader.java
@@ -340,11 +340,7 @@ final class GIFMetadataReader implements AutoCloseable {
 
     private byte[] read(int length) throws IOException {
         byte[] data = new byte[length];
-        int n, offset = 0;
-        while ((n = inputStream.read(
-                data, offset, data.length - offset)) < offset) {
-            offset += n;
-        }
+        inputStream.readFully(data);
         return data;
     }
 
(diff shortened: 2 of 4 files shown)


adam-vessey:
Test failures seem to be due to tests failing to pull in binaries to start the run? [`grok` seems to have deleted all the releases prior to 12 from GitHub](https://github.com/GrokImageCompression/grok/releases), and [the tests are trying to pull 7.something](https://github.com/cantaloupe-project/cantaloupe/blob/21057cf72071840e2a984450bf8845c6e6490ec9/docker/Linux-JDK11/Dockerfile#L31-L34).

And something off in the Windows build, it seems?

La0:
~~Any idea why some unit tests are not passing here ?~~

Just saw your previous comment afterward, so it may be unrelated to your changes

glenrobson:
I think when we merge in the changes in #656 or #653 we can update this branch and the tests should be fixed.

glenrobson:
Cantaloupe dev call 3rd July.

Thanks changes look good and will look again once the library updates have been applied to develop.

jcoyne:
@adam-vessey would you be able to rebase this on the develop branch?  That should allow all the tests to pass.  Then we can look at merging this.

adam-vessey:
Taking a cursory look through the failing tests, generally seems to be testing failure conditions, expecting `edu.illinois.library.cantaloupe.processor.SourceFormatException` exception, but now it is returning `java.io.EOFException`. `java.io.EOFException` is similarly a `java.io.IOException`, so unsure how the methods here quite had an effect on this? As in, am expecting to see something on the wrapping around the reads to consistently rethrow/wrap `java.io.IOException`s as `[...].SourceFormatException`s given the test assertion, but not seeing where it is? Or, are the test assertions too specific? 
Or yeah, should there be some try/catch wrapping? Probably in `JPEG2000MetadataReader.readData()` to catch the `java.io.IOException`s, to wrap with `[...].SourceFormatException`, with something in the message about image corruption?

---

As in, something like: https://github.com/adam-vessey/cantaloupe/compare/fix/unsafe-reads...adam-vessey:cantaloupe:fix/wrapped-reads ?

Created a draft PR with it, to see what happens: #681 

Seems to have done the trick over on #681 (not sure about whatever is happening with the linux/graalvm build; don't believe it's anything I did)

DiegoPino:
@adam-vessey merged! https://github.com/cantaloupe-project/cantaloupe/commit/dc7947a6d37f3282fc9c010afbd1a5f169ca5e06
Feel free to rebase. Thanks again

adam-vessey:
@DiegoPino : Closing this one in favour of #681, given the JPEG2000 metadata reader tests would still fail here without the changes being made there, or changes of some other nature (reworking the tests to more generic exceptions?).

PR #668, opened 2024-07-03
UCLA updates

glenrobson:
Taking edits from #656 and seeing if we can keep the new s3 funconality and XMP data


src/main/java/edu/illinois/library/cantaloupe/cache/S3MultipartAsyncOutputStream.java
```diff
@@ -229,7 +228,9 @@ public void run() {
     private boolean requestCreated;
 
     private String uploadID;
-    private int partIndex;
+    // Part number must start with 1
+    // minio will hang if this is 0
```
DiegoPino, line 232: @glenrobson just a comment. I think this is where AWS S3 specs v/s Minio's ones are different.

DiegoPino (approved):
@glenrobson congrats on all tests passing! I did a general sanity read/pass and all looks good to me. Approving

src/main/java/edu/illinois/library/cantaloupe/cache/S3MultipartAsyncOutputStream.java
glenrobson: I think AWS also requires parts to start at 1 but hopefully it wouldn't crash :-). https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPart.html

Issue #669, opened 2024-07-03
GrokImageCompression urls fail


jcoyne:
It's in all the Dockerfiles. Here is an exemplar:
https://github.com/cantaloupe-project/cantaloupe/blob/21057cf72071840e2a984450bf8845c6e6490ec9/docker/Linux-JDK11/Dockerfile#L31-L32

```
➜  curl https://github.com/GrokImageCompression/grok/releases/download/v7.6.5/libgrokj2k1_7.6.5-1_amd64.deb
Not Found%
➜  curl https://github.com/GrokImageCompression/grok/releases/download/v7.6.5/grokj2k-tools_7.6.5-1_amd64.deb
Not Found%
```

In the release branch we've commented this code out:
https://github.com/cantaloupe-project/cantaloupe/blob/release/5.0/docker/Linux-GraalVM20/Dockerfile#L35-L39

ksclarke:
Three months ago, the Grok project (https://github.com/GrokImageCompression/grok/) wiped its repository history and removed all old releases. They then released a new library version, and the repository was reset to a completely clean state. We asked about this change on the discussion board, and the question was closed. We asked about it, then, in a ticket/issue, and were told the project would not maintain its previous history or continue to host the old releases. Checking back on the project's repo now, I see all the old issues (including ours asking about the change) have also now been deleted and that the discussion board is no longer available.

UCLA made the decision that we were not comfortable relying on an upstream project with these practices and removed Grok from our Cantaloupe Docker image. The question for the larger Cantaloupe community would be whether we want to update its Dockerfile builds and documentation to reference the new version of Grok or remove it from Cantaloupe's list of supported libraries. UCLA's Docker image was reconfigured to use OpenJPEG (and Kakadu if the binaries are available). Still, I think this is a question for the Cantaloupe community to decide which path the Cantaloupe project should take. Are there people/institutions relying on Grok for its speed advantage over OpenJPEG (but not using Kakadu because of its proprietary license)?

Which path should Cantaloupe take?

mmatela:
> Are there people/institutions relying on Grok for its speed advantage over OpenJPEG (but not using Kakadu because of its proprietary license)?

I am one of those, so I would very much prefer for the Grok support to continue.
Do you really need to drop the support for Grok altogether? It's perfectly understandable if you don't include it in your docker files, but why not let us install it at our own risk?

jcoyne:
[...]

PR #670, opened 2024-07-03
Bugfix/xhr allow authorization header

ksclarke:
Retargeting the cantaloupe-project/cantaloupe#583 PR to the `develop` branch.


jcoyne:
Will this close #580?

ksclarke:
Yes, thanks, I've added that to main PR message b/c I'm not clear on whether that auto-close functionality will work from a comment.

Fwiw, CI build is failing because of the [Grok issue](https://github.com/cantaloupe-project/cantaloupe/issues/669).

ksclarke:
Kevin needs to look at this and resolve the broken CI builds. JDKs were updated but that's resulting in broken CI builds.

jcoyne (approved):

ksclarke:
> Will this close #580?

I was incorrect before saying "yes" to this. Mark was disagreeing with the way Alex implemented the fix and I need a bit more time to look into the differences. I've re-created a draft PR (#697) to compare. This current PR is okay as is, though.

src/main/java/edu/illinois/library/cantaloupe/resource/iiif/v1/InformationResource.java
```diff
@@ -42,6 +46,21 @@ public Method[] getSupportedMethods() {
         return SUPPORTED_METHODS;
     }
 
+    @Override
+    protected final void doOPTIONS() {
```
jcoyne, line 50: Is there a reason we can't put this method in `IIIFResource` and then we only need it one time rather than the three times?

jcoyne (changes requested):

src/main/java/edu/illinois/library/cantaloupe/resource/iiif/v1/InformationResource.java
ksclarke: I remember asking this question to Mark, too, but unfortunately do not remember what his response was at the time.

Just looking at it myself, there are other things that extend IIIF1Resource/IIIF2Resource/IIIF3Resource (which extend IIIFResource), like: LandingResource, IdentifierResource, and ImageResource. My guess would be that Mark didn't feel the doOptions() implementation he added was appropriate for those other resources, so kept the scope more limited to the info resources (across all 3 versions).

glenrobson: @jcoyne are you happy with this? Note the IIIF1Resource etc are the only info requests which need protecting so the other resources don't require auth.

ksclarke: A deeper refactoring could happen so that the headers were only presented on appropriate objects while reducing the duplication, but that would involve a more significant refactoring of Cantaloupe itself, which seems more involved than our first lower bar goal of just getting caught up on the queued PRs.

jcoyne (approved):

DiegoPino (approved):

PR #676, opened 2024-07-31
Include S3 object content type in attributes

garyttierney:
The `ContentTypeHeaderChecker` is invoked after the `NameChecker`, but the `contentType` is never set when the attributes are fetched from S3. This patch fixes that.


jcoyne:
@garyttierney Can you rebase this on the latest version of develop?  That ought to fix the test failures.

garyttierney:
@jcoyne yep, done!

glenrobson:
Glen to look at suggested issues:
 * Add specific imports back into the code
 * Now this is fixed what happens to the existing behaviour which may work on the file name extension 
 
 Glen to look into this issue.

glenrobson:
Closing in favour of https://github.com/cantaloupe-project/cantaloupe/pull/700

Issue #654, opened 2024-05-08
Review the JDK version support matrix for a next release/tests


DiegoPino:
# What?

Current tests include JDK 11, see https://github.com/cantaloupe-project/cantaloupe/blob/21057cf72071840e2a984450bf8845c6e6490ec9/.github/workflows/release.yml#L13
JDK11 is End of Life. 

With https://github.com/cantaloupe-project/cantaloupe/pull/653 coming in, for a 6.0.0 release we could probably drop old JDK versions and focus only on LTS/still valid versions of Java

see https://endoflife.date/oracle-jdk

glenrobson:
No objections recieved so far. Cantaloupe dev meeting on 3rd of July raised no objections to dropping JDK11.

glenrobson:
Which versions do we want to support?

 * 17 and 21
 * 22 looking to future 

Go by the support for OpenJDK: https://access.redhat.com/articles/1299013#OpenJDK_Life_Cycle

(which is much clearer than Oracles page).

mrghosti3:
Java v22 is a non-LTS release so not sure how suitable it is for the future compared to v21.

jcoyne:
@mrghosti3 I believe we were thinking "last two  LTS (17,21) + latest non-LTS (22)".  This means in September, when OpenJDK 23 is released, we can switch to 17,21,23.

mrghosti3:
Oh, understood. I guess I misunderstood the message. :sweat_smile:

PR #689, opened 2024-08-15
Test with OpenJDK 17 & 21 (LTS)

jcoyne:
Ref #654

### DIFF
--- a/docker/Linux-JDK17/Dockerfile
+++ b/docker/Linux-JDK17/Dockerfile
@@ -1,0 +1,56 @@
+FROM ubuntu:lunar
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+# Install various dependencies:
+# * ca-certificates is needed by wget
+# * ffmpeg is needed by FfmpegProcessor
+# * wget download stuffs in this dockerfile
+# * libopenjp2-tools is needed by OpenJpegProcessor
+# * All the rest is needed by GrokProcessor
+RUN apt-get update && apt-get install -y --no-install-recommends \
+  ca-certificates \
+  ffmpeg \
+  wget \
+  libopenjp2-tools \
+  liblcms2-dev \
+  libpng-dev \
+  libzstd-dev \
+  libtiff-dev \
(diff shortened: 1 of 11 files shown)


DiegoPino:
@jcoyne @glenrobson are we dropping 11? RE https://github.com/cantaloupe-project/cantaloupe/issues/654 ? Or it is a keeper?

jcoyne:
@DiegoPino I was planning on doing that separately to avoid tying the two changes together in case removal of 11 becomes contentious.

docker/Linux-JDK17/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+  libgrokj2k1 \
+  grokj2k-tools \
+  adduser \
+  openjdk-17-jdk \
+  maven \
+  && rm -rf /var/lib/apt/lists/*
+
+# Install TurboJpegProcessor dependencies
+RUN mkdir -p /opt/libjpeg-turbo/lib
+COPY docker/Linux-JDK11/image_files/libjpeg-turbo/lib64 /opt/libjpeg-turbo/lib
```
DiegoPino, line 33: We might want to move the built jpeg-turbo library into a shared space in case we deprecate JDK11. Just a note

docker/Linux-JDK21/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+WORKDIR $home
+
+# Install application dependencies
+COPY ./pom.xml pom.xml
+
+RUN echo "export JAVA_HOME=$(dirname $(dirname $(readlink -f $(which java))))" > ~/.bashrc
+RUN mvn --quiet dependency:resolve
+
+# Copy the code
+COPY --chown=cantaloupe docker/Linux-JDK11/image_files/test.properties test.properties
```
DiegoPino, line 53: same with the test.properties. Should be in a shared space instead of inside a specific JDK folder

DiegoPino (approved):
more tests!

PR #698, opened 2024-08-28
Stop testing on JDK11

jcoyne:
It is EOL in October 2024:
https://endoflife.date/redhat-build-of-openjdk


ksclarke (approved):
Looks good.

PR #699, opened 2024-09-11
Adding 5 retries for purge

### LINKED ISSUE #678: S3CacheTest.testPurge:330 is a flaky test
This test seems to fail randomly on CI:
```
cantaloupe_1  | [ERROR]   S3CacheTest.testPurge:330 expected: <false> but was: <true>
```

glenrobson:
Not a very sophisticated fix but it looks like the purge doesn't have a back door to check when the process has finished. This change introduces 5 retries before failling. 

Closes #678


DiegoPino (approved):
Great! 5 is a lucky number

PR #700, opened 2024-09-11
Include S3 object content type in attributes

glenrobson:
The `ContentTypeHeaderChecker` is invoked after the `NameChecker`, but the `contentType` is never set when the attributes are fetched from S3. This patch fixes that.

Imported branch from #676 and added specific imports back.


glenrobson:
Question from previous meeting:

> Now this is fixed what happens to the existing behaviour which may work on the file name extension

According to the [documentation](https://github.com/cantaloupe-project/cantaloupe/blob/0cc9057572c21b4601fe1d93b569f2b67d085c3e/src/main/java/edu/illinois/library/cantaloupe/source/S3Source.java#L93) (and a brief look at the [code](https://github.com/cantaloupe-project/cantaloupe/blob/0cc9057572c21b4601fe1d93b569f2b67d085c3e/src/main/java/edu/illinois/library/cantaloupe/source/S3Source.java#L161)) it looks like the process is:
```
 * <ol>
     *     <li>If the object key has a recognized filename extension, the
     *     format is inferred from that.</li>
     *     <li>Otherwise, if the source image's URI identifier has a recognized
     *     filename extension, the format will be inferred from that.</li>
     *     <li>Otherwise, a {@literal GET} request will be sent with a
     *     {@literal Range} header specifying a small range of data from the
     *     beginning of the resource.
     *         <ol>
     *             <li>If a {@literal Content-Type} header is present in the
     *             response, and is specific enough (i.e. not {@literal
     *             application/octet-stream}), a format will be inferred from
     *             that.</li>
     *             <li>Otherwise, a format is inferred from the magic bytes in
     *             the response body.</li>
     *         </ol>
     *     </li>
     * </ol>
 ```

So fixing this issue will only change the behaviour if the format can not be inferred from the filename extension.

**Comment by glenrobson** [comment]
Some more information here:

https://github.com/cantaloupe-project/cantaloupe/blob/f492eac72d3d01e46d6dad911cb124a401ce86cb/src/main/java/edu/illinois/library/cantaloupe/source/Source.java#L71

```
/**
     * <p>Returns an iterator over the results of various techniques of
     * checking the format, in the order of least to most expensive. Any of the
     * calls to {@link Iterator#next()} or may return either an inaccurate
     * value, or {@link Format#UNKNOWN}. Clients should proceed using the first
     * non-unknown format they encounter and, if this turns out to be wrong,
     * iterate and try again.</p>
     *
     * @return Iterator over whatever format-inference strategies the instance
     *         supports. <strong>The instance is cached and the same one is
     *         returned every time.</strong>
     * @since 5.0
     */
    Iterator<Format> getFormatIterator();
```

ksclarke (approved):

PR #701, opened 2024-09-25
Add web identity token file credentials provider to credentials chain

glenrobson:
Moving fixes to point to develop branch


glenrobson:
Allows the setting of a enviromental variable `aws.webIdentityTokenFile` to point to the web identity token file.


https://docs.aws.amazon.com/sdkref/latest/guide/feature-assume-role-credentials.html

glenrobson:
Add documentation for using this new facility. 

Look at logging as this can generate a lot of logs if the first set authentication doesn't pass.

Issue #706, opened 2024-10-16
"Too many files open" during API requests to clear cache


bbpennel:
We have an automated workflow that makes requests to the ["PurgeItemFromCache" API endpoint](https://cantaloupe-project.github.io/manual/5.0/endpoints.html#PurgeItemFromCache) in cantaloupe 5.0.5 after a jp2 gets replaced in our system, so that all of its derivatives in cantaloupe's cache will get regenerated. But we've had cantaloupe run out of file handles repeatedly while regenerating a set of 5700 jp2s.

I'm able to replicate the issue pretty easily in our system by making concurrent calls to "PurgeItemFromCache" with a populated cache. If I monitor the file handles held by the process, I can watch them quickly pile up and never be released. They are basically all references to empty directories, like:
```
lr-x------ 1 root cantaloupe 64 Oct 16 11:50 100 -> /opt/cantaloupe-5.0.5/cache/image/06/6c/55
lr-x------ 1 root cantaloupe 64 Oct 16 11:50 1000 -> /opt/cantaloupe-5.0.5/cache/image/a4/49/f7
lr-x------ 1 root cantaloupe 64 Oct 16 11:50 1001 -> /opt/cantaloupe-5.0.5/cache/image/a4/49/f7
lr-x------ 1 root cantaloupe 64 Oct 16 11:50 1002 -> /opt/cantaloupe-5.0.5/cache/image/af/d2/05
lr-x------ 1 root cantaloupe 64 Oct 16 11:50 1003 -> /opt/cantaloupe-5.0.5/cache/image/af/d2/05
lr-x------ 1 root cantaloupe 64 Oct 16 11:50 1004 -> /opt/cantaloupe-5.0.5/cache/image/af/d2/05
lr-x------ 1 root cantaloupe 64 Oct 16 11:50 1005 -> /opt/cantaloupe-5.0.5/cache/image/af/d2/05
```

I don't know if the same issue occurs with a request to purge the whole cache, or when the background cache clearing processes run. This is occurring in a RHEL7 server. I don't see any similar behaviors come from read requests to cantaloupe, like requesting tiles or thumbnails.

bbpennel:
I've opened a PR which seems to fix the issue on our local servers. I've done several hundred purge requests without any new file handles being added.

DiegoPino:
Fixed via https://github.com/cantaloupe-project/cantaloupe/commit/0dccf299b9310f2293e6cb467799a5345fd58e39
[...]

PR #707, opened 2024-10-16
Prevent file handle leak during individual purge

bbpennel:
https://github.com/cantaloupe-project/cantaloupe/issues/706

Use a try-with-resource block when getting the stream of files in derivative directory to avoid file handle leak

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/cache/FilesystemCache.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/cache/FilesystemCache.java
@@ -508,8 +508,8 @@ class FilesystemCache implements SourceCache, DerivativeCache {
                 hashedPathFragment(identifier.toString()));
         final String expectedNamePrefix =
                 StringUtils.md5(identifier.toString());
-        try {
-            return Files.list(cachePath)
+        try (final var fileStream = Files.list(cachePath)) {
+            return fileStream
                     .filter(p -> p.getFileName().toString().startsWith(expectedNamePrefix))
                     .collect(Collectors.toUnmodifiableSet());
         } catch (NoSuchFileException e) {


src/main/java/edu/illinois/library/cantaloupe/cache/FilesystemCache.java
```diff
@@ -508,8 +508,8 @@ Set<Path> getDerivativeImageFiles(Identifier identifier)
                 hashedPathFragment(identifier.toString()));
         final String expectedNamePrefix =
                 StringUtils.md5(identifier.toString());
-        try {
-            return Files.list(cachePath)
+        try (final var fileStream = Files.list(cachePath)) {
+            return fileStream
                     .filter(p -> p.getFileName().toString().startsWith(expectedNamePrefix))
```
DiegoPino, line 513: @bbpennel. Nice! so the try() will auto close the fileStream... But... is the filter needed? Given the way caches are structured all files inside cachePath will need to be returned. Maybe for sanity just check if the file is regular file?
.filter(`Files::isRegularFile`) or something like that? I mean, just to save a few milliseconds per cycle... specially on large images with tons of derivatives

glenrobson:
Cantaloupe group met and like this fix. If you can respond to the filter question reaised above then we can look at merging.

src/main/java/edu/illinois/library/cantaloupe/cache/FilesystemCache.java
bbpennel: The filter isn't a change in this PR, i didn't really evaluate whether it was needed or not. I'm fine with removing it if there isn't any concern of the cache path containing files that don't begin with the md5.

DiegoPino: @bbpennel thanks, true!, for the sake of not adding any point of rupture and because I don't know what use case the original Developer might had had in place, let's keep it! Approving. Thanks!

DiegoPino (approved):

DiegoPino:
@bbpennel there is a test failing (looks related... but might be yet again a transient/timeout error on writing a file/reading it)
JdbcCacheTest>AbstractCacheTest.testPurgeInfos:469->AbstractCacheTest.assertExists:36 expected: not <null>
So will re-run this test if that is the case

Issue #680, opened 2024-08-01
Incorrect upscaling for some files with size=max


mmatela:
Using cantaloupe 5.0.6, default settings except for FilesystemSource.BasicLookupStrategy.path_prefix. Java 21.0.1 on Windows, but experienced the same on other versions of Java on Linux.

Attaching sample file 2.jpg ![2](https://github.com/user-attachments/assets/eca73bfc-7194-4a50-9830-beee020d9e1a)

When trying to get a fragment of size 100x100 with this url: http://localhost:8182/iiif/3/2.jpg/100,100,100,100/max/0/default.jpg i get the fragment upscaled to size 2604x2604 instead.

It works correctly when using size pct:100 instead of max (should there ever be a difference for these two?): http://localhost:8182/iiif/3/2.jpg/100,100,100,100/pct:100/0/default.jpg
It works correctly with image api v2: http://localhost:8182/iiif/2/2.jpg/100,100,100,100/full/0/default.jpg
Somehow it accepts size=max for v2, and also returns upscaled to 2604x2604: http://localhost:8182/iiif/2/2.jpg/100,100,100,100/max/0/default.jpg

I tried some other files in both jpg and other formats and for most of them it correctly returns 100x100, so it must be something inside the content file, but I'm not aware of any properties that should lead to this result.

DiegoPino:
Hi @mmatela. Could you share the info.json output for that image? 
If this was a general issue for all the images I would say there is a problem/order of operation with the Specs implementation for `/max` but since you reported it happens just in some images it might be something specific maybe to A) wrongly processed by cantaloupe resolution (your image has 400dpi) ? wrongly parsed Exif? or even stale caches?. Still the specs are clear.
`Region THEN Size THEN Rotation THEN Quality THEN Format
`
For the stale caches part (this is just an idea, not stating really a solution but worth trying) you can call the tile URL with
```
Overriding Cache Behavior
The behavior of the [caches](https://cantaloupe-project.github.io/manual/5.0/caching.html) can be overridden on a per-request basis by supplying a cache URL query argument with one of the following values:

nocache or false
Bypasses the source, derivative, and info caches, and also omits any Cache-Control response header that may be configured.
recache
Bypasses the derivative and info caches for reads but not writes.
```

I will check the code that sets up the operations and see what I can find out

mmatela:
Thanks for looking into this @DiegoPino 

[...]

PR #709, opened 2024-11-06
ISSUE-680: proper /MAX handling under pre-cropped and un-cropped scenarios

DiegoPino:
# What?

Solves #680 

I tried to be as minimalistic (code-wise) as possible (after some iterations that had nothing of minimal).

# HOW?

Basically now, before processing `MAX` at `::constrainSizeToMaxPixels`, we request from the Operation List the actual "So far what size would this image result in?", and use that `Dimension` as base for all calculations. This way if a `REGION/CROP` (if any) results in less pixel area than `MAX_PIXELS`, `MAX` has no effect, but if not, scaling is done proportional to the REGION requested (including if `REGION\CROP` exceeded the actual source dimensions .. which is IIIF valid), and will then constrain the final dimensions to fit into a MAX_PIXELS sized region.

As always, test code is > than the actual fix :)

Note: according to me this works (tested with #680 image too). Test pass. But would be nice though if someone had the time to give this a human test.


DiegoPino:
I'm getting Tests failing here (not related to this pull? -- yes, related to this pull).

`Version3_0ConformanceTest.testXYRegionOutOfBounds:206 expected: <400> but was: <500>`

Fixing this

DiegoPino:
FYI: This assumes there is max. a single Crop (or extending class) operation in the Operation List. I don't know if somehow one could via a normal processing chain add more than one at all? but maybe (via code?) someone would?

jcoyne (approved):

glenrobson (approved):

Issue #712, opened 2024-11-11
slow memory leak in ConcurrentLocalContextProvider with ruby delegate


matthias-fratz-bsz:
Cantaloupe appears to be slowly leaking memory in `ConcurrentLocalContextProvider` (screenshot from Eclipse MAT on a heap dump):

![ConcurrentLocalContextProvider retaining about 8GB](https://github.com/user-attachments/assets/8cd36f60-62f4-4f21-b8ad-1f7d4c6ae5b8)

That class seems to be caching the per-thread JRuby contexts. Whenever a thread terminates, its context becomes useless and needs to be disposed of. As far as I can tell from the code, this only happens at application shutdown, when `terminate()` walks the `contextRefs` list and cleans up all the JRuby contexts.

Jetty with its (default?) `EatWhatYouKill` threading strategy does create and destroy threads. (At the point of the heap dump, the server had gone through ~66000 threads.) Thus the `contextRefs` queue keeps filling up until the process eventually runs out of memory. The heap dump seems to confirm this, as it shows the ~8GB being retained by `contextRefs`.

This is by no means a massive leak: 8GB over 66k threads is only around 100kB per thread. But as our server runs for weeks without restarts, that does add up. IMO it would be better to dispose of the per-thread contexts using a background / cleanup thread, activated whenever a thread is terminated, without waiting until process termination.

I came across this when troubleshooting random hangs in Cantaloupe (similar to #238); the heap dump I used is from an unresponsive instance. I don't know how the memory leak could cause a hang, but not having found anything else out of the ordinary, it is my best guess for what might be causing it. At 8GB, the process *is* at its heap limit, though strangely no OutOfMemoryExceptios are being thrown.

matthias-fratz-bsz:
We are experiencing the issue on Cantaloupe 5.0.6, but the relevant code is in the JRuby library anyway. JRuby 9.4.3.0 (what Cantaloupe uses as of 0dccf299), and in fact also JRuby latest, contains code that I would expect to exhibit this leaking behavior.

I could offer to fix the issue, but I don't know whether this really is a bug in `ConcurrentLocalContextProvider`. Maybe that class is simply being used outside its specified use case in Cantaloupe, and isn't supposed to be kept around for a long time? I don't really feel qualified to make that distinction...

I see that Cantaloupe does specifically select that context provider in `JRubyDelegateProxy`:

```
// Available values are singleton, singlethread, threadsafe and
// concurrent (JSR-223 default). See
[...]

PR #715, opened 2024-12-11
disable JRuby's Variable Sharing feature

matthias-fratz-bsz:
This avoids a memory leak in JRuby 9.4.3.0 (probably at least 9.4.x).

If Variable Sharing is enabled, JRuby stores all Ruby instance variables into a `BiVariableMap` object for access from Java. As of the current 9.4.10.0 development version, this `BiVariable` object is never removed from its `BiVariableMap`, and thus never gets garbage-collected. `JRubyDelegateProxy` creates a new instance of the Ruby `CustomDelegate` object, so any use of instance variables in that object leaks a bit of memory after each request. This eventually grinds Cantaloupe to a halt.

Disabling Variable Sharing is a workaround until this can be fixed in JRuby. JRuby's Variable Sharing feature isn't needed by the code in `JRubyDelegateProxy`.

This memory leak is similar to the one described in #712, but the mechanism is slightly different. I'm proposing the workaround because according to synthetic tests, it should also reduce the impact of that other leak quite significantly.

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/delegate/JRubyDelegateProxy.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/delegate/JRubyDelegateProxy.java
@@ -6,6 +6,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.script.Invocable;
+import javax.script.ScriptContext;
 import javax.script.ScriptEngine;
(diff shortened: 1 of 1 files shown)


glenrobson:
Note related issue: https://github.com/jruby/jruby/issues/8527

glenrobson (approved):

PR #731, opened 2025-02-13
Fix the issue of 404 after upgrading Jetty12

zzmloveddm:
I have tried to fix the 404 issue encountered, and it can start normally and process images. After deployment, it passed the security department's detection, but I did not test other logic. You can try to see if other parts are working properly


DiegoPino:
@zzmloveddm thanks so much for your contribution! I will review this, test locally and if all goes well merge against the jetty-12 branch. Tests won't pass yet (I committed to refactor those tomorrow) so please hang tight before trying to run this branch on production. Appreciate it!

Issue #724, opened 2025-01-29
Wrong Sequence extraction on Video at the Image Resource Level


DiegoPino:
# What?

There are some consistency issues when forming the INFO object across different type of media. Single images generate a single source, PDFs one for each Page, but Videos, even if they allow a "second" sequence request, only generate a single Source. Because the ImageResource is unaware of this, under e.g second 2, there will be call asking for the second image in the INFO resource throwing an exception.

Because the Image Resource Class / and that specific method has little access to any other variables than the actual INFO object, one hawkish way of solving that would be to add a helper (or just a check) that uses the INFO resource to check if a resource is a VIDEO, and there, independently of the passed sequence ID, we always request KEY 0 (second 1).

This is bug 

@patdunlavey I waited for your issue, but we have a call today, so I went and opened it myself

DiegoPino:
Resolved via https://github.com/cantaloupe-project/cantaloupe/commit/cf22dace43b44d5e216c3a2c0409af49c013ad8e

PR #738, opened 2025-02-26
ISSUE-724: Fixes access to frames other than 1 for Video

DiegoPino:
See #724 

The way I approached is:
- least intrusive/effort (to avoid modifying the N+1 places we access getSize(pageIndex) all over the place
- Simplest to test by tapping into the Info Class itself, which is not V1/V2/V3 specific like all the other resources

Added a few basic tests. Funny enough, our resource tests don't really use the Info class many times ... but I hope this covers it 

FYI: this works

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/image/Info.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/image/Info.java
@@ -407,7 +407,12 @@ public final class Info {
      * @return Size of the image at the given index.
      */
     public Dimension getSize(int imageIndex) {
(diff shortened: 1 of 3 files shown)


DiegoPino:
@glenrobson @jcoyne @ksclarke could any of you review my code style/sanity for this one? I feel it is a bit hack-ish so happy to address any concerns. All tests are passing now

jcoyne (approved):
Looks good.

glenrobson (approved):

PR #759, opened 2025-04-10
Allow spaces in the `format` argument of `FormatException`

### LINKED ISSUE #757: Unsupported format message is rendered without spaces
When I request an image in a file format that is not supported for the original, I get a "415 Unsupported Media Type" error with a strange-looking message: "Unsupported format: UnsupportedformatAvailableoutputformatsforthisimagearelistedintheinformationresponse".
The example is asking for a WebM movie for a still image. I do expect the error type, but not the no-space message.

I think the logic in

https://github.com/cantaloupe-project/cantaloupe/blob/fac584f55bd26f783fb4ebd02c97b82038fdcc54/src/main/java/edu/illinois/library/cantaloupe/resource/iiif/FormatException.java#L5-L7

does not work well with the way the error is constructed in

https://github.com/cantaloupe-project/cantaloupe/blob/fac584f55bd26f783fb4ebd02c97b82038fdcc54/src/main/java/edu/illinois/library/cantaloupe/resource/iiif/v3/Parameters.java#L107-L113

(and similarly in the v2 `Parameters` class).

This is a minor issue, but I wanted to share it anyway. I assume the reason to also remove spaces in the `FormatException` constructor, is that a `FormatException` was meant to be created with the requested format as the argument. However, it looks like the only direct use of the constructor is with the above message. 

On the other hand, when I change the request to ask for WebP, which we also don't support, the message I get *does* say: "Unsupported output format: WebP". Wait! I see there are two different `FormatException` classes.

I think we can safely change the `format.replaceAll("[^A-Za-z0-9]", "")` because *that* code is only called with the message I saw for my WebM request. My WebP request was handled in an `OutputFormatException`.

bencomp:
This closes #757. However, I'm open to discuss whether this is the way to go in the issue comments.

When a IIIF v1 endpoint receives a request for an image in an unsupported format, the requested format (i.e. the filename extension) is provided as the `format` argument in a `FormatException`.
When a IIIF v2 or v3 endpoint receives a request for an unsupported format, i.e. a format that is not in the registry, the `format` argument is a help message – a suggestion to look for available formats in the information resource.
The `format` is normalised/cleaned and the result is used to construct a more elaborate error message.

This PR changes the normalisation to allow spaces (but not `.`s that are also in the help messages for v2 and v3) and fixes those help messages by adding a missing space.

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/resource/iiif/FormatException.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/resource/iiif/FormatException.java
@@ -3,7 +3,7 @@ package edu.illinois.library.cantaloupe.resource.iiif;
 public class FormatException extends IllegalArgumentException {
 
     public FormatException(String format) {
-        super("Unsupported format: " + format.replaceAll("[^A-Za-z0-9]", ""));
+        super("Unsupported format: " + format.replaceAll("[^A-Za-z0-9 ]", ""));
     }
 
 }
--- a/src/main/java/edu/illinois/library/cantaloupe/resource/iiif/FormatException.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/resource/iiif/FormatException.java
@@ -3,7 +3,7 @@ package edu.illinois.library.cantaloupe.resource.iiif;
 public class FormatException extends IllegalArgumentException {
 
     public FormatException(String format) {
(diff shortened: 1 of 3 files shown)


DiegoPino (approved):

DiegoPino:
@bencomp we discussed and we will go with this once tests pass (thanks) and open a different one for normalizing. We really appreciate your contribution

Issue #330, opened 2019-11-08
JaiProcessor availability


kiktestdmp:
I've read that the processor is quite old and no longer being developed. But will cantaloupe continue to incorporate this processor as an option ?

The reason I ask, is because it is incredibly good to the point where the other processors for Pyramidal Tiff to not even compare. The difference in performance is like having internet connection speed 1 mbit or 500 mbit ...

We've decided to use cantaloupe as our institutional IIIF server, but the reason is mostly due to cantaloupe offering this JaiProcessor making it much faster than all other IIIF servers for TIFF. 

If JaiProcessor were to be removed, I would need around 6-7 loadbalanced servers running IIIF instances to have something near the performance of 1 server with jaiprocessor.

PR #763, opened 2025-04-11
[6.x] Remove Deprecated JAIUtil and JaiProcessor

jcoyne:


glenrobson:
See this issue from 2019:

https://github.com/cantaloupe-project/cantaloupe/issues/330

Would be useful to do some speed and memory testing with pyramid tiffs to see the preformance hit if we dropped JaiProcessor.

Issue #751, opened 2025-03-20
Upgrade TurboJPEG (JNI Bindings and calling code) to 3.x


DiegoPino:
# What?

This is a macro issue. our JNI bindings for TurboJpeg support 2.x. Some users have reported using TurboJpeg 3.x without success (500 errors) regarding mismatch of native C v/s the JAVA wrappers.

- Probably the first part to look at are the Bindings https://github.com/libjpeg-turbo/libjpeg-turbo/tree/main/java/org/libjpegturbo/turbojpeg
- Then test/adapt (if needed) our own code
- Document TurboJpeg 3.x being the new "normal" 

Se #749 and #750

DiegoPino:
@glenrobson Will assign this to myself bc I don't see any takers!

glenrobson:
If you are busy I could look at it next week. I'm getting more familiar with bindings after the kakadu work :-)

DiegoPino:
Yeah, a bit busy lately. If you find space before I make a pull, please re-assign to you! Thanks @glenrobson

glenrobson:
OK Ill take a look early next week.

jcoyne:
It doesn't look like Ubuntu 20-24 ship with libjpeg-turbo 3. That has to be installed with `apt get libjpeg-turbo-official`.  Might it be more complicate if we link against a non-official package?

DiegoPino:
True. So you are all saying this is a future ubuntu problem (next LTS release is April 2026?)
Same goes for 
- Debian LTS. Current 11 ships with 2. https://packages.debian.org/bullseye/source/libjpeg-turbo
- For Alpine Linux (if you are building smaller docker containers) latest does provide 3.x (r0) version https://pkgs.alpinelinux.org/packages?name=libjpeg*&branch=v3.21&repo=&arch=x86_64&origin=&flagged=&maintainer=
But not sure they have the LTS idea here (3.13 ships with 2.x)

glenrobson:
So should the task here be to test libjpeg-turbo 2.1 and include it in the docker files? If so I think the actions are:

- [ ] Update the docker files in https://github.com/cantaloupe-project/cantaloupe/tree/develop/docker to use the system version of turbojpeg
- [ ] Update the bindings in https://github.com/libjpeg-turbo/libjpeg-turbo/tree/main/java/org/libjpegturbo/turbojpeg (I'm hoping that v2 bindings will work with all 2.x versions of libjpeg-turbo but we'll see)
[...]

Issue #749, opened 2025-03-19
Does TurboJpegProcessor support libjpeg-turbo greater than 2.0.2 - releases 3.x and up?


hrvoj3e:
What is the official response from core team?

https://cantaloupe-project.github.io/manual/5.0/processors.html#TurboJpegProcessor
> This processor requires libjpeg-turbo 2.0.2 to be installed.

But 2.0.2 is from Feb 14, 2019, and Ubuntu repo offers 2.1.0 from Nov 2021 and current release is 3.0.1 from Dec 2024.

Also, there is a official [APT/YUM repo](https://libjpeg-turbo.org/Downloads/YUM) for new releases. Can we use that or is there braking changes inside JNI ?

Also, what is the version of  libjpeg-turbo used in [cantaloupe/docker/image_files/libjpeg-turbo
/lib64/](https://github.com/cantaloupe-project/cantaloupe/tree/9ef196eca34b2ca648abbe0b365324a7aedd5aaf/docker/image_files/libjpeg-turbo/lib64) ?

DiegoPino:
@hrvoj3e as it stands right now, you might have luck with other 2.x but 3.x is not working.

hrvoj3e:
Out of curiosity - can I ask why? 
Has 3.x many breaking changes or it is not worth it (performance/feature wise)?

I suppose they have resolved many security,memory issues in those 3-4 years.


btw. we have tried 3.x and it works on images but not with out test PDF.

DiegoPino:
Well, the nature of software release is that on every increment there might be improvements and bug fixes, that is clear and the ideal dream in OSS is to support every dependency as it evolves.

But also, we inherited Cantaloupe not long ago and the efforts on the last release were to bring the base system up to date (including security wise) and fix breaking IIIF API contracts, but updating everything to every latest version requires time and resources, and thorough testing, documentation and also awareness that moving forward might leave others behind (or make their upgrade paths more complex). Semantically (as in semantic versioning) the last release was a minor release (in the 5.x) so meant to still be compatible and hopefully a drop In replacement for existing users. Not saying we are always successful and code/errors can happen. Also part of the nature of OSS.

 We are always happy to accept pull requests and contributions and for sure 3.x will be on our radar now for a next release. The "right now" is not a "we don't support in the future" statement.

[...]

PR #777, opened 2025-07-02
Using system turbo jpeg

### LINKED ISSUE #760: Upgrade TurboJPEG (JNI Bindings and calling code) to latest 2.1 version
Transfered from https://github.com/cantaloupe-project/cantaloupe/issues/751

- [x] Update the docker files in https://github.com/cantaloupe-project/cantaloupe/tree/develop/docker to use the system version of turbojpeg
- [x] Update the bindings in https://github.com/libjpeg-turbo/libjpeg-turbo/tree/main/java/org/libjpegturbo/turbojpeg (I'm hoping that v2 bindings will work with all 2.x versions of libjpeg-turbo but we'll see)
- [x] Remove the binary files from https://github.com/cantaloupe-project/cantaloupe/tree/develop/docker/image_files/libjpeg-turbo/lib64

Due to turbo jpeg 2.1 being distributed with Ubuntu we are sticking with that version until 3.x becomes more prevalent.

glenrobson:
So should the task here be to test libjpeg-turbo 2.1 and include it in the docker files? If so I think the actions are:

- [x] Update the docker files in https://github.com/cantaloupe-project/cantaloupe/tree/develop/docker to use the system version of turbojpeg
- [x] Update the bindings in https://github.com/libjpeg-turbo/libjpeg-turbo/tree/main/java/org/libjpegturbo/turbojpeg (I'm hoping that v2 bindings will work with all 2.x versions of libjpeg-turbo but we'll see)
- [x] Remove the binary files from https://github.com/cantaloupe-project/cantaloupe/tree/develop/docker/image_files/libjpeg-turbo/lib64

For 3.0 I think we should comment on #749 to say we will get to it when its distrubted in core repostiories.  

From https://github.com/cantaloupe-project/cantaloupe/issues/751#issuecomment-2741709324


glenrobson:
Note on Ubutnu Noble the libjpeg version is: Version: 1:2.1.5-2ubuntu2

jcoyne (approved):

PR #780, opened 2025-07-17
Remove deprecated illegal-access argument

jcoyne:
This no longer has an effect in Java 17+


glenrobson (commented):
Looks good and now we only support Java 17+ it would be good to remove to hide the warning.

glenrobson (approved):

PR #783, opened 2025-07-17
Upgrade jruby to 10.0.0.1

### LINKED ISSUE #712: slow memory leak in ConcurrentLocalContextProvider with ruby delegate

jcoyne:


glenrobson:
Requires Java 22+ which means we need to remove JDK 17 which will happen when the next LTS comes out.

glenrobson (approved):

PR #784, opened 2025-07-17
Fix adduser syntax

jcoyne:


glenrobson:
No objection to merging this but just wondering why this change? Don't they do the same thing?

glenrobson:
Trying to fix issue in build log:

```
#11 0.197 Use of uninitialized value $answer in pattern match (m//) at /usr/sbin/adduser line 879.
```


https://github.com/cantaloupe-project/cantaloupe/actions/runs/16348237581/job/46187684455

jcoyne:
```
adduser --home $home $user
``` 
Does seem like the right syntax, but we do seem to have the problem.

PR #801, opened 2025-10-06
Object#finalize is deprecated and shouldn't be called

jcoyne:


src/main/java/org/libjpegturbo/turbojpeg/TJCompressor.java
```diff
@@ -582,8 +582,6 @@ protected void finalize() throws Throwable {
     try {
```
glenrobson, line 582: It looks like its not a good idea to have a finalize method at all. Is it worth removing it?

glenrobson: From https://www.baeldung.com/java-finalize

src/main/java/org/libjpegturbo/turbojpeg/TJDecompressor.java
```diff
@@ -869,8 +869,6 @@ protected void finalize() throws Throwable {
     try {
```
glenrobson, line 869: Again should we remove this hole method?

glenrobson:
Could these files implement AutoCloseable?

DiegoPino:
Autocloseables @glenrobson sound like a more convenient idea than depending on the garbage collector to call finalize and close() -- if that even runs/before we run out of resources.

PR #816, opened 2025-10-08
Simplify ErrorResource

### LINKED ISSUE #813: ErrorResource should not extend AbstractResource.
We probably don't need a complex inheritance hierarchy here.

jcoyne:
Namely, make it inherit nothing, because there is no natural inheritance relationship.

Fixes #813

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/resource/HandlerServlet.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/resource/HandlerServlet.java
@@ -142,21 +142,13 @@ public class HandlerServlet extends HttpServlet {
                              HttpServletResponse response,
                              Throwable t) {
         // Try to use an ErrorResource, which will render an HTML template.
-        ErrorResource resource = new ErrorResource(t);
+        ErrorResource resource = new ErrorResource(t, request, response);
         try {
-            // N.B.: the response status will be set by ErrorResource based on
-            // the type of Throwable.
-            response.setContentType("text/html;charset=UTF-8");
-            resource.setRequest(new Request(request));
-            resource.setResponse(response);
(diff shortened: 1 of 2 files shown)


src/main/java/edu/illinois/library/cantaloupe/resource/HandlerServlet.java
```diff
@@ -142,21 +142,13 @@ private void handleError(HttpServletRequest request,
...
         // Try to use an ErrorResource, which will render an HTML template.
-        ErrorResource resource = new ErrorResource(t);
+        ErrorResource resource = new ErrorResource(t, request, response);
         try {
-            // N.B.: the response status will be set by ErrorResource based on
-            // the type of Throwable.
-            response.setContentType("text/html;charset=UTF-8");
-            resource.setRequest(new Request(request));
-            resource.setResponse(response);
-            resource.doInit();
```
glenrobson, line 152: Just noting we loose the X-Powered-By header which is set by the doInit

https://github.com/cantaloupe-project/cantaloupe/blob/999370f4cc7fe06684b5cc073131f357dddccd0f/src/main/java/edu/illinois/library/cantaloupe/resource/AbstractResource.java#L150

glenrobson (approved):

src/main/java/edu/illinois/library/cantaloupe/resource/HandlerServlet.java
jcoyne: Thanks. I've restored it here: https://github.com/cantaloupe-project/cantaloupe/pull/816/files#diff-d837e07e9ead48a4ab4a090f0ab25719673d25c56b82bcf8c70afb45011a1575R103

PR #817, opened 2025-10-08
Extract a ContentTypeNegotiator from AbstractResource

jcoyne:


src/main/java/edu/illinois/library/cantaloupe/resource/AbstractResource.java
```diff
@@ -698,19 +659,6 @@ public Method[] getSupportedMethods() {
...
     }
 
-    /**
-     * @param limitToTypes Media types to limit the result to, in order of most
-     *                     to least preferred by the application.
-     * @return             Best media type conforming to client preferences as
-     *                     expressed in the {@code Accept} header; or {@code
-     *                     null} if negotiation failed.
-     */
-    protected final String negotiateContentType(List<String> limitToTypes) {
```
glenrobson, line 708: Is this used by subclasses?

glenrobson: I see it is below :-)

glenrobson (approved):

PR #825, opened 2025-10-08
Fix content negotiation

jcoyne:
The Allow header is not for a list of methods. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Allow

When 405 is returned, then Allow is required to be provided and be a list of methods: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/405


glenrobson:
I think Allow is allowed for an OPTIONS request: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/OPTIONS#examples

Issue #814, opened 2025-10-08
Flatten AbstractResource > PublicResource > IIIIFResource hierarchy


jcoyne:
Why do we need it? Maybe blocked by #813

glenrobson:
Note only three classes extended IIIF Resource:

https://github.com/search?q=repo%3Acantaloupe-project%2Fcantaloupe%20IIIFResource&type=code

with very minimal funconality.

PR #827, opened 2025-10-08
Merge IIIFResource and PublicResouce

jcoyne:
there is no reason to keep these two classes separate

Ref #814


glenrobson:
It looks like IIIFResource used to extend AbstractResource until this point:

https://github.com/cantaloupe-project/cantaloupe/commit/51a639b913422eb6c3fde5691f4b7e075c6c7d85#diff-c0b43c6cf5b57756193edfe7157b19868230519fe242c3fe1f0bfd211d2909ef

but I agree with only IIIFResource extending PublicResource it seems unnessary.

glenrobson (approved):

PR #831, opened 2025-10-09
Extract the Basic Auth method from AbstractResource

jcoyne:
The AbstractResource has too many responsibilities.  This will enable us to move to Spring Boot.

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/auth/BasicAuth.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/auth/BasicAuth.java
@@ -1,0 +1,47 @@
+package edu.illinois.library.cantaloupe.auth;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+
+import edu.illinois.library.cantaloupe.http.Status;
+import edu.illinois.library.cantaloupe.resource.ResourceException;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+
+public class BasicAuth {
+        /**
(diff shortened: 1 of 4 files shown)


src/main/java/edu/illinois/library/cantaloupe/auth/BasicAuth.java
```diff
@@ -0,0 +1,47 @@
...
+     * @throws ResourceException if authentication failed.
+     */
+    public static final void authenticateUsingBasic(String realm,
+                                                CredentialStore credentialStore,
+                                                HttpServletRequest request,
+                                                HttpServletResponse response)
+            throws ResourceException {
+        boolean isAuthenticated = false;
+        String header = request.getHeader("Authorization");
+        if (header != null && "Basic ".equals(header.substring(0, Math.min(header.length(), 6)))) {
```
glenrobson, line 28: Any reason not to use String.startsWith?

glenrobson (approved):
One question above but not a major concern for me.

src/main/java/edu/illinois/library/cantaloupe/auth/BasicAuth.java
jcoyne: No, just my ignorance of the standard libraries.

jcoyne: This was just a move of code from https://github.com/cantaloupe-project/cantaloupe/pull/831/files#diff-34c2b754ed0fd057e1c2eb63adaf0034cbeee3f506d8a74771a8a51b3b08be8aL252

PR #834, opened 2025-10-09
Remove unnecessary method getMethod()

jcoyne:
This is only ever used for logging.


glenrobson:
I think its used here:      

https://github.com/cantaloupe-project/cantaloupe/blob/4e3f46c50cd029fca27758a3698c66a721457d8c/src/main/java/edu/illinois/library/cantaloupe/resource/HandlerServlet.java#L87

jcoyne:
@glenrobson I think `request` on that line is an instance of `HttpServletRequest`, not `edu.illinois.library.cantaloupe.resource.Request`

glenrobson (approved):
I think you are right about the  request being a HttpServletRequest.

Changes look good to me.

PR #838, opened 2025-10-09
Fix redirect to align with the documentation

jcoyne:
Previously it was redirecting to a full URI, but the documentation of the method indicates it should redirect to a path.

See documentation:
https://github.com/cantaloupe-project/cantaloupe/blob/646e89b14639b4f1daaa322f5bc40aa42cb00b4a/src/main/java/edu/illinois/library/cantaloupe/resource/TrailingSlashRemovingResource.java#L10-L11

### DIFF
--- a/src/test/java/edu/illinois/library/cantaloupe/resource/iiif/v3/LandingResourceTest.java
+++ b/src/test/java/edu/illinois/library/cantaloupe/resource/iiif/v3/LandingResourceTest.java
@@ -45,8 +45,7 @@ class LandingResourceTest extends ResourceTest {
 
     @Test
     void testGETWithTrailingSlashRedirectsToWithout() throws Exception {
-        final URI uri = getHTTPURI("");
-        assertRedirect(new URI(uri + "/"), uri, 301);
+        assertRedirect(getHTTPURI("/"), Route.IIIF_3_PATH, 301);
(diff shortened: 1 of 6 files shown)


src/test/java/edu/illinois/library/cantaloupe/resource/iiif/v3/LandingResourceTest.java
```diff
@@ -3,6 +3,7 @@
 import edu.illinois.library.cantaloupe.Application;
 import edu.illinois.library.cantaloupe.config.Configuration;
 import edu.illinois.library.cantaloupe.config.Key;
+import edu.illinois.library.cantaloupe.http.Client;
```
glenrobson: Is this used anywhere?

glenrobson:
Should this be a v6 thing? I know we are matching the documentation but if people have setup their server expecting the full server and path redirect it might break things if we release it in a v5 update...

src/test/java/edu/illinois/library/cantaloupe/resource/iiif/v3/LandingResourceTest.java
jcoyne: no. I've removed it.

PR #844, opened 2025-10-10
Remove the builder pattern in favor of a simple constructor

jcoyne:


glenrobson (changes requested):
Initally I was thinking this looked good as it reduced complexity but comparing some of the before and after code it looks to me less readable. For example:

```
try (InformationRequestHandler handler = InformationRequestHandler.builder()
                .withCallback(callback)
                .withIdentifier(new Identifier("jpg-rgb-64x48x8.jpg"))
                .withRequestContext(context)
                .build()) 
```

Reads better than:

```
try (InformationRequestHandler handler = new InformationRequestHandler(
                new Identifier("jpg-rgb-64x48x8.jpg"),
                null,
                new RequestContext(),
                callback,
                false,
                false)) {
```
I think its also losing some of the validation e.g:

```
 if (handler.identifier == null) {
                throw new IllegalArgumentException("Identifier cannot be null.");
            } else if (handler.delegateProxy != null &&
                    handler.requestContext == null) {
                throw new IllegalArgumentException("If a delegate proxy is " +
                        "set, a request context must also be set.");
            }
            if (handler.requestContext == null) {
                handler.requestContext = new RequestContext();
            }
            return handler;
```

jcoyne:
@glenrobson The problem is that using the builder pattern could allow the user to create an object that is in an invalid state.  When you use the builder pattern, every step should be valid.   In this case it's not a suitable choice. I think you are right that it's less readable, but that's more of a bug of the Java language not having keyword arguments.


I tossed the validation, because we are always passing a Identifier object, never a variable which could be null-able.  This is just code that is never called.

glenrobson (approved):

Issue #848, opened 2025-10-10
Changes to the architecture of a web request


jcoyne:
Currently we create a `edu.illinois.library.cantaloupe.resource.Request` object for every HTTP request.

I posit, that the `Request` class ought to have an inheritance structure:
```
Request - For any HTTP request, deals with things like proxy urls.
-> IIIF request - Holds version endpoint (e.g. `/iiif/v1`)
  -> IIIF With ID -> Things that have an identifer ( Identifier, Information, and Image)
    -> IIIF With transform -> For the Image requests.
```

Currently we sort of have this hierarchy, (e.g. `AbstractResource` -> `IIIFResource` -> `IIIF2Resource` -> `v2.InformationResource`). However the classes in the current hierarchy contains both the model of what the user is requesting and the controller code that deals with HTTP idioms.   In the new architecture the `Request` hierarchy is modeling the users intent, and we will have controllers that deal with implementing that request using the HTTP idioms.

PR #849, opened 2025-10-10
Move the identifier methods into the Request object

jcoyne:
Move them out of the AbstractResource

Ref #848

### DIFF
--- a/docker/Linux-JDK25/Dockerfile
+++ b/docker/Linux-JDK25/Dockerfile
@@ -38,14 +38,15 @@ ENV LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu:${LD_LIBRARY_PATH}
 
 # Install various other dependencies that aren't in apt
 # Install OpenJDK
+#https://download.java.net/java/GA/jdk25.0.2/b1e0dfa218384cb9959bdcb897162d4e/10/GPL/openjdk-25.0.2_linux-x64_bin.tar.gz
 RUN wget -q https://download.java.net/java/GA/jdk25.0.2/b1e0dfa218384cb9959bdcb897162d4e/10/GPL/openjdk-25.0.2_linux-x64_bin.tar.gz \
   && tar xfz openjdk-25.0.2_linux-x64_bin.tar.gz \
   && mv jdk-25.0.2 /opt/jdk \
   # Install a newer Maven than the one in apt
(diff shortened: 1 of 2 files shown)


src/main/java/edu/illinois/library/cantaloupe/resource/Route.java
```diff
@@ -7,6 +7,8 @@
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.checkerframework.checker.units.qual.g;
```
glenrobson, line 10: What does this class do? Seems to represent mass i.e. grams....

src/main/java/edu/illinois/library/cantaloupe/resource/Route.java
```diff
@@ -130,15 +152,22 @@ List<String> getPathArguments() {
         return pathArguments;
     }
 
+    Route(Class<? extends AbstractResource> resource, Class<? extends Request> request) {
```
glenrobson, line 155: Should this constructor be higher up in the class? i.e. before the instance methods, maybe below the static methods? 

I know you are copying whats there but is there any reason these aren't public methods?

src/main/java/edu/illinois/library/cantaloupe/resource/iiif/SizeConstrainer.java
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * When the size expressed in the endpoint URI is {@code max}, and the
+     * resulting image dimensions are larger than {@link Key#MAX_PIXELS}, the
+     * image must be downscaled to fit that area.
+     * 
+     * @param requestedSize  Full size of the source image.
+     * @param opList OperationsList. This gets mutated if the requested size is too big
+     * @throws ValidationException if a cropping Operation is invalid.
+     */
+    public static void constrainSizeToMaxPixels(Dimension requestedSize,
```
glenrobson, line 24: Where is this used? Is it being setup for another change down the line?

glenrobson:
Need to merge: 

https://github.com/cantaloupe-project/cantaloupe/pull/841

and 

https://github.com/cantaloupe-project/cantaloupe/pull/840

first as these contain the size and scale constraingts.

jcoyne:
We need #890 first

PR #861, opened 2025-10-22
Update S3Source.region comment

jcoyne:
We found that using a non AWS endpoint, and omitting the region caused loading files to take about 20 seconds.  Adding a dummy value makes loading faster. Potentially due to some code in the aws-sdk.


glenrobson:
Group agrees to merge this.

DiegoPino (approved):

Issue #870, opened 2025-11-04
JP2K with OpenJpegProcessor and AutomaticSelectionStrategy


vljukap98:
It seems as if my OpenJpeg processor works only when `processor.selection_strategy=ManualSelectionStrategy` and `processor.ManualSelectionStrategy.jp2=OpenJpegProcessor`.

When my `cantaloupe.properties` defines `processor.selection_strategy=AutomaticSelectionStrategy` and I try to request a jp2 image with OpenJPEG installed on my system (or via docker). The cantaloupe processing fails, actual server log doesn't show anything interesting, but the stack trace is actually included in the failed response:

```
500 Internal Server Error



java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:74)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:483)
	at edu.illinois.library.cantaloupe.processor.ProcessorFactory.newProcessor(ProcessorFactory.java:100)
	at edu.illinois.library.cantaloupe.resource.ImageRequestHandler.handle(ImageRequestHandler.java:385)
	at edu.illinois.library.cantaloupe.resource.iiif.v2.ImageResource.doGET(ImageResource.java:141)
	at edu.illinois.library.cantaloupe.resource.HandlerServlet.handle(HandlerServlet.java:97)
	at edu.illinois.library.cantaloupe.resource.HandlerServlet.doGet(HandlerServlet.java:35)
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:500)
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:587)
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:764)
	at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:529)
	at org.eclipse.jetty.server.handler.ScopedHandler.nextHandle(ScopedHandler.java:221)
	at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1381)
	at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:176)
	at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:484)
[...]

PR #872, opened 2025-11-04
ISSUE-870: Deals with kakadu libs not present (same way in the past TurboJpeg not present was dealt with)

DiegoPino:
See #870 

Adds an implicit constructor to Kakadu, catches `UnsatisfiedLinkError` but ignores it so 
https://github.com/ksclarke/cantaloupe/blob/78028a21b1058655c35039339ea6bbec32eba8a8/src/main/java/edu/illinois/library/cantaloupe/processor/ProcessorFactory.java#L40 can do its job, discarding it as a candidate if the libs are not present.

This effectively moves "on" to OpenJPEG if automatic selection is enabled. Basically does what done before on TurboJpeg https://github.com/cantaloupe-project/cantaloupe/commit/8508e9353be4597153fff8298d8b43508808c76c

@glenrobson @jcoyne might need your "Coding standards" & logic review on both code and wording here please

@vljukap98 would you mind giving this a try? Hope this helps


jcoyne:
This seems reasonable to me.

vljukap98:
@DiegoPino I tried it and it looks good

DiegoPino:
@jcoyne @vljukap98 @glenrobson because automatic testing is not running and all the machines I have around (like easy access ... I could spine one next week) are arm64 (so not intel) I can't run the test with actually kakadu enabled. My fear is that  I could break something when it is enabled with this pull.. so, anyone could, when time permits, run `mvn` with tests and share the output? Even maybe like test manually with aJP2image if kakadu still works? Appreciate it

glenrobson (approved):

PR #878, opened 2025-12-02
Spring boot 4

jcoyne:


pom.xml
```diff
@@ -310,7 +311,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-annotations</artifactId>
-            <version>${jackson.version}</version>
+            <version>${jackson-annotations.version}</version>
```
jcoyne: jackson-annotations no longer uses a patch level version component https://mvnrepository.com/artifact/com.fasterxml.jackson.core/jackson-annotations

PR #882, opened 2025-12-17
Increase wait time for JdbcCache test

jcoyne:
There are async operations afoot.  After `purgeInfos()` is called in the test, any pending `purgeInfoAsync()` tasks in the `TaskQueue` could still execute and interfere with the test assertions. The 3.5-second `ASYNC_WAIT` may not be sufficient in slower CI environments.


DiegoPino (approved):
Looks good.

Issue #879, opened 2025-12-10
LZW tiffs corrupted when not cropped


dfdan:
I've noticed that on both 5.0.7 and current develop HEAD, TIFFs with LZW (and possibly other?) compression methods are corrupted when requesting /full/ region, while cropping avoids this. This is gives strange results for thumbnails / initial image loads in viewers, which magically corrects itself when the user zooms in. 

Example: 

uncompressed works fine:
.../uncomp.tiff/full/200,/0/default.jpg:

<img width="200" height="200" alt="Image" src="https://github.com/user-attachments/assets/d4b7c53b-f4b1-490a-90ae-a407d721b633" />

however, compressed version of same image is corrupted:
.../comp.tiff/full/200,/0/default.jpg:

<img width="200" height="200" alt="Image" src="https://github.com/user-attachments/assets/46647a48-ad5f-49c0-a2c5-9e4a4e69733f" />

however cropping this works:
.../comp.tiff/0,0,500,500/200,/0/default.jpg

<img width="200" height="200" alt="Image" src="https://github.com/user-attachments/assets/5f454f15-8f65-402b-9928-0d36bc5b0d63" />

Some quick poking suggests that images are opened with different methods depending on whether a crop being obtained or not, here: https://github.com/cantaloupe-project/cantaloupe/blob/b519614bd93352b41b852851c8702fd1235988b7/src/main/java/edu/illinois/library/cantaloupe/processor/codec/AbstractIIOImageReader.java#L393

I assume that `iioReader` is somehow not handling compressed images correctly? 

Simple test files attached. I initially saw this with real/in the wild images.

[comp.tif](https://github.com/user-attachments/files/24075078/comp.tif)
[uncomp.tif](https://github.com/user-attachments/files/24075077/uncomp.tif)

glenrobson:
Thank you for submitting this issue and providing the test files. We will look into testing this issue and creating a fix.

DiegoPino:
@dfdan hi. I did a first and very quick test with the compressed TIFF under a slightly different than 5.0.7 version I have running (see the difference [here](https://github.com/cantaloupe-project/cantaloupe/compare/89faf9f..55c60b9)) and I saw no corruption or error of any type. 

[...]

PR #885, opened 2025-12-19
ISSUE-879: Revert "Upgrade to latest version of imageio-ext-tiff"

DiegoPino:
Reverts cantaloupe-project/cantaloupe#809. See #879 

Gist is: imageio-ext-tiff 14+ (and still an issue in 2.x) introduced changes in the way the LZW decompressing is happening. Until we figure out an alternative or it gets fixed upstream, staying at 13 seems to be the safest path

@jcoyne @glenrobson please merge at your discretion, thanks

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -153,7 +153,11 @@
         <dependency>
             <groupId>it.geosolutions.imageio-ext</groupId>
             <artifactId>imageio-ext-tiff</artifactId>
-            <!-- This seemed latest version 1.3.2: https://github.com/geosolutions-it/imageio-ext 2023 -->
+            <!-- Delay upgrading to 1.4.x or 2.x.x until 
+                 https://github.com/geosolutions-it/imageio-ext/issues/346 is resolved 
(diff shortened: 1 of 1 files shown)


jcoyne (approved):

pom.xml
```diff
@@ -153,7 +153,8 @@
         <dependency>
             <groupId>it.geosolutions.imageio-ext</groupId>
             <artifactId>imageio-ext-tiff</artifactId>
-            <version>1.4.14</version>
+            <!-- This seemed latest version 1.3.2: https://github.com/geosolutions-it/imageio-ext 2023 -->
```
jcoyne: How about adding a note here explaining that 1.4.14 had this bug, and a link to the upstream issue?

DiegoPino: @jcoyne I added a message there. let me know how that reads. Thanks!

jcoyne (approved):

pom.xml
jcoyne: This is fantastic.

Issue #875, opened 2025-11-20
o.a.j.riot [ErrorHandlerFactory.java:120] [line: 1, col: 58608] The prefix "Camera" for element "Camera:Trait" is not bound.


tohruok:
Summary: Pixel JPEG XMP may contain elements (e.g. Camera:Trait, GImage:MicroVideo) without namespace declarations. In Util.readModel(), any RiotException other than the base-URI case is rethrown, aborting merge and losing XMP.

Steps to Reproduce:

Obtain Pixel-generated JPEG.
Extract APP1 XMP segments and call assembleXMP().
readModel() throws RiotException: prefix "... not bound".
Current Behavior: Parsing aborts; extended / standard merge fails; metadata lost.

Expected: Graceful repair of common undeclared Google prefixes (Camera, GImage, GPano) or selective skipping while retaining remaining triples.

Proposed Change: Pre-scan rdfXML and inject missing known namespaces before parsing. Limit catch scope; log repaired cases.

Patch sketch:

edu.illinois.library.cantaloupe.processor.codec.jpeg.Util.java

        // ...existing code...
        private static Model readModel(String rdfXML) {
            Model model = ModelFactory.createDefaultModel();
            String base = null;

            // Heuristic: inject missing Google Pixel XMP namespaces
            if (rdfXML != null) {
                if (rdfXML.contains("Camera:") && !rdfXML.contains("xmlns:Camera=")) {
                    rdfXML = rdfXML.replaceFirst("<rdf:RDF",
                            "<rdf:RDF xmlns:Camera=\"http://ns.google.com/photos/1.0/camera/\"");
                }
        
            }

            try (StringReader reader = new StringReader(rdfXML)) {
                model.read(reader, base, "RDF/XML");
            } catch (RiotException e) {

         // ...existing code...

glenrobson:
We aren't currently looking at fixing metadata in images as this could lead to a lot of edge cases and would encourage the fixing of XMP metadata in images before using cantaloupe.

We are looking at making the XMP processing more robust to log issues and try and carry on rather than breaking. 

If you can supply the full Exception stack trace we can make sure we put a exception catch in the correct place so the image processing can contiune.

tohruok:
Thank you for the feedback. I understand the approach - making XMP processing more robust by logging issues rather than breaking the image processing flow makes sense.

Here is the full exception stack trace:

````
[...]

PR #886, opened 2025-12-22
ignore RiotException: #875

TohruOkamoto:


TohruOkamoto:
JDBC cache tests are not related to the scope of the fix

glenrobson:
We need to investigate the ramifications on returning an empty model if it fails to parse the XMP headers. We do see a lot of XMP parsing issues so it would be worth looking into this. 

Thank you for submitting the pull request.

glenrobson:
Action: test this change with the open XMP issues.

src/main/java/edu/illinois/library/cantaloupe/processor/codec/jpeg/Util.java
```diff
@@ -121,17 +125,22 @@ private static String mergeXMPModels(String standardXMP,
...
+            try (StringReader reader = new StringReader(rdfXML)) {
+                model.read(reader, base, "RDF/XML");
+            } catch (RiotException exception) {
+                if (exception.getMessage().indexOf("Base URI is null, but there are relative URIs to resolve") != -1) {
+                    // Version 4.8+ of jena requires a rdf:about link to not be empty
+                    try (StringReader reader = new StringReader(rdfXML)) {
+                        model.read(reader, "http://example.com", "RDF/XML");
+                    }    
+                } else {
+                    throw exception;
```
jcoyne: Can we put the LOGGER here and not have the outer try/catch block?

glenrobson: Rather than catching the exception again you could replace line 138 with:

```
LOGGER.warn(ex.getMessage(),ex);
// Return empty model on parse failure.
```

glenrobson (changes requested):

src/main/java/edu/illinois/library/cantaloupe/processor/codec/jpeg/Util.java
TohruOkamoto: As you pointed out, we fixed the handling of try catch.

PR #892, opened 2026-02-09
fix: Keep S3 endpoint override from config when using basic strategy

renanmedina:
## 📋 Overview

This PR fixes the S3 source behavior when using the **basic** lookup strategy so that a configured endpoint override (`S3source.endpoint`) is applied correctly. It ensures that when the basic strategy is used, the resolved endpoint from configuration is set on `S3ObjectInfo`, matching the behavior of the delegate strategy and keeping client caching and resolution consistent.

Also keeps errors consistent with proper information:

### Before
image

### After
image

### 1. Endpoint override for basic strategy
- In `getObjectInfoUsingBasicStrategy()`, the endpoint is now read from configuration (`Key.S3SOURCE_ENDPOINT`) and set on the returned `S3ObjectInfo` via `info.setEndpoint(endpoint)`.
- Previously, only the delegate strategy set the endpoint on the object info; the basic strategy did not, which could lead to inconsistent behavior or incorrect client resolution when a custom endpoint (e.g. MinIO or another S3-compatible store) is configured.

### 2. Debug logging
- Added debug log statements to indicate which lookup strategy was used (delegate vs basic) and the resulting object info.
- Added a final debug log for the object info returned by `getObjectInfo()` to aid troubleshooting.

## 📝 Technical Details

| File | Description |
|------|-------------|
| `src/main/java/edu/illinois/library/cantaloupe/source/S3Source.java` | In `getObjectInfoUsingBasicStrategy()`, read `S3SOURCE_ENDPOINT` from config and call `info.setEndpoint(endpoint)` so the basic strategy sets the endpoint on `S3ObjectInfo`. Added debug logging in `getObjectInfo()` and in both delegate and basic strategy branches. |

## ✅ Testing

- [x] With **basic** lookup strategy and `S3source.endpoint` set (e.g. MinIO or custom S3-compatible endpoint), verify image requests resolve and stream correctly.
- [x] With **basic** strategy and no endpoint override (default AWS S3), verify behavior is unchanged.
- [x] With **delegate script** lookup strategy, verify behavior is unchanged (endpoint still comes from delegate result).
- [x] Optionally enable debug logging and confirm the new strategy/object-info debug messages appear as expected.


glenrobson:
Looks a good change. Just need to work out why we haven't come across this problem in production with working with s3.

jcoyne:
I think the issue is on line 321, where getClientInstance is not using the endpoint within S3ObjectInfo.

renanmedina:
@glenrobson since it was not setting the custom endpoint the lookup probably always defaulted to the null index of clients. If the application always have a default client created that should be the reason why it does not happen in production (i did not track the entire application behavior).

I just made sure to keep it consistent since I saw the delegate branch was always setting the endpoint as well and wanted to make clear when debugging the correct behavior.

thanks for having the time to read this PR ;)

PR #897, opened 2026-02-12
Upgrade to jetty 12.1.3

jcoyne:
See https://github.com/jetty/jetty.project/releases/tag/jetty-12.1.3

This is the last version of jetty 12 that we can upgrade to without upgrading slf4j. See #902


DiegoPino:
@jcoyne can we close this one? Since you have a newer one open?

jcoyne:
@DiegoPino yes, if we merge #902. Otherwise, if we don't like #902, then this is a better option than "no change".

DiegoPino:
Closing bc #902 was merged. Thanks!

PR #910, opened 2026-03-18
Remove finalize() from kdu JNI classes

jcoyne:
Finalize is deprecated by Java


jcoyne:
This is generated code and shouldn't be modified.

Issue #912, opened 2026-04-06
GrokProcessor hangs for some files


mmatela:
I'm running cantaloupe-5.0.7.jar with `processor.ManualSelectionStrategy.jp2 = GrokProcessor`. I've installed grok2jk 20.2.6 via brew.sh.
Example file (changed the extension to make sure it's easy to dwonload without github reconverting the image): 
[image.jp2.dmp](https://github.com/user-attachments/files/26510605/image.jp2.dmp)

When requesting for example http://127.0.0.1:8182/iiif/3/image.jp2/0,0,4096,4096/512,/0/default.jpg all the bytes of the resulting image are sent, but it looks like no EOF is sent so the client waits for further chunks indefinitely.
I see that during this time there's a child process spawned with command like `/home/linuxbrew/.linuxbrew/bin/grk_decompress -i /tmp/test/image.jp2 -r 2 -o /tmp/GrokProcessor-f4aef21e-5370-4317-8933-0e3ffc81408f.bmp`. But Grok seems to be working fine, because invoking the same command manually (with different output file) completes immediately.

There's no problem with most other jp2 files generate with the same method (`grk_compress -t 512,512 -r 20 -i "input.tif" -o image.jp2`.

There's no problem with this file when I switch to `OpenJpegProcessor`.

jcoyne:
As we discussed in https://github.com/cantaloupe-project/cantaloupe/issues/669#issuecomment-2263227701, we don't currently have a maintainer that is invested in Grok support, but you have spoken up to keep it in the release, which we appreciated!

Unfortunately, this means we may not be able to offer much help diagnosing this particular issue. All the same, thank you for filing this report, as it may be helpful for anyone else encountering this problem.  If you're able to investigate further and discover what's causing this, we'd be very happy to review a PR. And please do share if you find a workaround!

mmatela:
I'm happy to report that one of AIs managed to come up with a plausible explanation of the problem and propose a one-line solution: https://aistudio.google.com/app/prompts/19SW0Usq6Dh-kSNIzofw7gB4eLCDij7yU

I checked out tag 5.0.7, added this line and compiled - it solved the problem in my setup!

In case there's a problem with the link, in `GrokProcessor.processInUnix` add line `processInputStream.transferTo(OutputStream.nullOutputStream());` before the line containing `image = Java2DPostProcessor.postProcess(`.

jcoyne:
[...]

PR #913, opened 2026-04-07
Fix potential deadlock in GrokProcessor #912

mmatela:
Fix for https://github.com/cantaloupe-project/cantaloupe/issues/912

# Explanation by AI:

Based on the thread dump, source code, and symptoms you provided, you've run into a classic **Java Process pipe deadlock**. 

Here is exactly what is happening and how to fix it.

### The Root Cause

1. **The Architecture**: In Unix, `GrokProcessor` tells `grk_decompress` to output the converted `.bmp` file to a symlink pointing to `/dev/stdout`. Java captures this via `process.getInputStream()`.
2. **The Read**: Java uses `BMPImageReader` to read the image. `BMPImageReader` is efficient—it reads the BMP header, figures out the dimensions, reads the exact bytes for the pixel array, and **stops reading immediately**.
3. **The Leftover Data**: Newer versions of Grok (`grokj2k`) often extract the ICC color profile or other metadata from the JP2 and embed it at the *end* of the resulting BMP file.
4. **The Deadlock**: When `BMPImageReader` stops reading, any trailing metadata (like the ICC profile) remains unread in the OS pipe buffer (which is typically 64KB on Linux). If this trailing data is larger than 64KB, the pipe fills up. `grk_decompress` gets blocked on a `write()` system call and cannot exit.
5. **The Hang**: 
    - Java successfully processes the parsed image and flushes it to the client (which is why you receive all the image bytes).
    - Java then reaches line 507: `final int code = process.waitFor();`
    - Because `grk_decompress` is blocked trying to write the rest of the metadata to a full pipe, it never exits. `waitFor()` blocks forever.
    - Since the handler method never returns, the Jetty server never sends the HTTP EOF chunk, and your client hangs indefinitely.

This explains why it works manually (writing to disk doesn't block like a pipe), why it works for most images (those images either lack an ICC profile or it's small enough to fit in the 64KB pipe buffer), and why `OpenJpegProcessor` works (it likely doesn't append the ICC profile to the BMP).

### The Fix

To solve this, we must ensure that Java drains the remainder of the process's standard output before calling `process.waitFor()`. This will unblock `grk_decompress` so it can exit cleanly.

You will need to patch the `GrokProcessor.java` file. Locate the `processInUnix` method and add the stream drain command right after `reader.read(0)`:

```java
// Around line 497 in GrokProcessor.java

BufferedImage image = reader.read(0);

// --- ADD THESE LINES TO FIX THE DEADLOCK ---
// Drain any remaining bytes (e.g. ICC profiles) from the stdout pipe.
// If we don't, the OS pipe buffer fills up, grk_decompress hangs,
// and process.waitFor() blocks indefinitely.
processInputStream.transferTo(OutputStream.nullOutputStream());
// -------------------------------------------

image = Java2DPostProcessor.postProcess(
        image, hints, opList, info, reductionFactor);

ImageWriterFacade.write(image,
        (Encode) opList.getFirst(Encode.class),
        outputStream);

final int code = process.waitFor();
```

*(Note: `transferTo(OutputStream.nullOutputStream())` is available because Cantaloupe 5.x requires Java 11+).*

By silently consuming and discarding the remaining bytes in the pipe, `grk_decompress` is allowed to finish its write operation, exit with code 0, and allow `process.waitFor()` to continue smoothly.

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/processor/GrokProcessor.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/processor/GrokProcessor.java
@@ -497,6 +497,11 @@ class GrokProcessor  extends AbstractProcessor implements FileProcessor {
                         EnumSet.of(ReaderHint.ALREADY_CROPPED);
 
                 BufferedImage image = reader.read(0);
+
+                // Drain any remaining bytes (e.g. ICC profiles) from the stdout pipe.
+                // If we don't, the OS pipe buffer may fill up and hang grk_decompress
+                processInputStream.transferTo(OutputStream.nullOutputStream());
+
                 image = Java2DPostProcessor.postProcess(
                         image, hints, opList, info, reductionFactor);
 


DiegoPino:
My concern: 
https://github.com/cantaloupe-project/cantaloupe/issues/912#issuecomment-4202854496

jcoyne:
Don't the comments in this PR disagree with the comments up in the head matter? https://github.com/cantaloupe-project/cantaloupe/blob/develop/src/main/java/edu/illinois/library/cantaloupe/processor/GrokProcessor.java#L57-L61

glenrobson:
Thank you @mmatela for submitting this pull request and raising the issue. As discussed in https://github.com/cantaloupe-project/cantaloupe/issues/912#issuecomment-4255545272 we are going to fix the original issue in hopefully a better way which will improve our support of colour spaces.

PR #916, opened 2026-04-13
Update github actions to latest versions

jcoyne:
actions/create-release has been archived. https://github.com/actions/create-release

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,38 +8,21 @@ jobs:
     name: Upload Release Asset
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/setup-java@v1
+      - uses: actions/setup-java@v5
         with:
+          distribution: "temurin"
           java-version: "21"
           java-package: jdk
           architecture: x64
       - name: Checkout Code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v6
       - name: Build Project
         # we assume that the tests for this tag have already passed
         # (otherwise the tag wouldn't have been pushed)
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -1,45 +1,28 @@
...
   build:
     name: Upload Release Asset
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/setup-java@v1
+      - uses: actions/setup-java@v5
         with:
-          java-version: '17'
+          distribution: "temurin"
+          java-version: "17"
```
glenrobson: Should this be 21?

.github/workflows/release.yml
```diff
@@ -1,45 +1,28 @@
...
       - name: Build Project
         # we assume that the tests for this tag have already passed
         # (otherwise the tag wouldn't have been pushed)
         run: mvn clean package -DskipTests
-      - name: Create Release
-        id: create_release
-        uses: actions/create-release@v1.0.0
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Create Release and Upload Asset
```
glenrobson, line 23: This looks good. Its much simpler than the old way and I did something similar in: https://github.com/IIIF-Commons/iiif_3d_blender_plugin/blob/main/.github/workflows/release_build.yml

.github/workflows/release.yml
jcoyne: That's in a different pr.

DiegoPino (approved):
Looks good to me

PR #917, opened 2026-04-14
Replace Minio with RustFS

jcoyne:
The minio repository has been archived


src/main/java/edu/illinois/library/cantaloupe/cache/S3Cache.java
```diff
@@ -132,7 +132,7 @@ public Optional<Info> getInfo(Identifier identifier) throws IOException {
                 .build();
         final Stopwatch watch = new Stopwatch();
         try (ResponseInputStream<GetObjectResponse> is = client.getObject(request)) {
-            // This extra validity check may be needed with minio server
+            // This extra validity check may be needed with rustfs server
```
DiegoPino, line 135: Is this still valid with rustfs @jcoyne ?

DiegoPino:
@jcoyne this looks good. We don't need to persist /data inside rustfs so a mount is not needed.

src/main/java/edu/illinois/library/cantaloupe/cache/S3Cache.java
jcoyne: I don't know.

glenrobson:
Closing this one in favour of Seaweed FS

PR #919, opened 2026-04-15
Replace Minio with SeaweedFS

jcoyne:


glenrobson:
Go with seaweed fs as a test proxy for s3.

getFormatIteratorNext test 

keep method but skip Format.UNKNOWN as s3 always requires a mime type. 

Remove this test: https://github.com/cantaloupe-project/cantaloupe/blob/f9335ac084dc39fff181582cb08a9a058c35e96f/src/test/java/edu/illinois/library/cantaloupe/cache/S3CacheTest.java#L249 

as its only testing minio.

DiegoPino (approved):
Looks good!

PR #923, opened 2026-04-22
S3 async CRT client

jcoyne:
For high throughput data transfer https://aws.amazon.com/blogs/storage/improving-amazon-s3-throughput-for-the-aws-cli-and-boto3-with-the-aws-common-runtime/


glenrobson:
Add cache s3 handling as well.

src/main/java/edu/illinois/library/cantaloupe/source/S3Source.java
```diff
@@ -7,12 +7,13 @@
 import edu.illinois.library.cantaloupe.image.Identifier;
 import edu.illinois.library.cantaloupe.image.MediaType;
 import edu.illinois.library.cantaloupe.delegate.DelegateMethod;
-import edu.illinois.library.cantaloupe.util.S3ClientBuilder;
```
glenrobson, line 10: Can we remove S3ClientBuilder class when we do the cache?

glenrobson (approved):

PR #928, opened 2026-05-01
Move the identifier methods into the Request object

jcoyne:
Move them out of the AbstractResource. An AbstractResource is something that is on the server, but these methods are about something in the user's Request.  These are fundamentally different objects.
Ref #848

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/resource/Request.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/resource/Request.java
@@ -6,6 +6,7 @@ import edu.illinois.library.cantaloupe.delegate.DelegateProxy;
 import edu.illinois.library.cantaloupe.http.Headers;
 import edu.illinois.library.cantaloupe.http.Query;
 import edu.illinois.library.cantaloupe.http.Reference;
+import edu.illinois.library.cantaloupe.image.Identifier;
 import edu.illinois.library.cantaloupe.image.MetaIdentifier;
 import edu.illinois.library.cantaloupe.util.StringUtils;
 import jakarta.servlet.http.HttpServletRequest;
@@ -26,6 +27,12 @@ public final class Request {
     private HttpServletRequest wrappedRequest;
     private Headers headers;
     private Reference reference;
+    private List<String> pathArguments;
+
+    /**
+     * Cached by {@link #getIdentifier()}.
+     */
+    private Identifier identifier;
 
     private static final Logger LOGGER =
             LoggerFactory.getLogger(Request.class);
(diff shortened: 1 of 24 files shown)


src/main/java/edu/illinois/library/cantaloupe/resource/iiif/v1/InformationResource.java
```diff
@@ -134,12 +134,12 @@ private void addHeaders(Info info, Information iiifInfo) {
 
     /**
      * @return Image URI corresponding to the given identifier, respecting the
-     *         {@code X-Forwarded-*} and {@link #PUBLIC_IDENTIFIER_HEADER}
+     *         {@code X-Forwarded-*} and {@link Request#PUBLIC_IDENTIFIER_HEADER}
      *         reverse proxy headers.
      */
     private String getImageURI() {
         return getRequest().getPublicRootReference() + Route.IIIF_1_PATH + "/" +
```
glenrobson, line 139: Is it worth turning this into a request method? Something like:

request.getImageURI(String versionPath) {
  retrun getPublicRootReference() + versionPath + getPublicIdentifier();
}

then it would be called:

return getRequest().getImageURI(Route.IIIF_1_PATH)

src/main/java/edu/illinois/library/cantaloupe/resource/iiif/v2/IdentifierResource.java
```diff
@@ -33,7 +33,7 @@ public void doGET() {
         final Reference newRef = new Reference(
                 getRequest().getPublicRootReference() +
                 Route.IIIF_2_PATH +
-                "/" + getPublicIdentifier() +
+                "/" + getRequest().getPublicIdentifier() +
```
glenrobson, line 36: if getRequest().getImageURI is implemented you could also add 

request.getImageInfo(versionPath) {
    return getImageURI(versionPath) + "/info.json";
}

src/main/java/edu/illinois/library/cantaloupe/resource/Request.java
```diff
@@ -43,15 +50,86 @@ public final class Request {
     /**
      * @param request Request that the new instance will wrap.
      */
-    Request(HttpServletRequest request) {
+    Request(HttpServletRequest request, List<String> pathArguments) {
```
glenrobson, line 50: with all the tests below is it worth creating another Constructor that only takes the request?

glenrobson: so instead of:

```
instance = new Request(servletRequest, Collections.emptyList());
```

you could use

```
instance = new Request(servletRequest);
```

and in request have:
```
Request(HttpServerletRequest request) {
      this(request, Collections.emptyList());
}
```

src/main/java/edu/illinois/library/cantaloupe/resource/Request.java
```diff
@@ -22,13 +23,19 @@
 /**
  * Wraps an {@link HttpServletRequest}, adding some convenience methods.
  */
-public final class Request {
```
glenrobson, line 24: Question whether we revert this change?

jcoyne, line 24: revert this.  There are no subclasses.

src/main/java/edu/illinois/library/cantaloupe/resource/Request.java
jcoyne: I'd prefer not to create special paths through the code just for testing.  The tests should exercise the paths that we want to use in production.

src/main/java/edu/illinois/library/cantaloupe/resource/iiif/v2/IdentifierResource.java
jcoyne: Possibly in the future. With this change I'm just trying to move things out of the AbstractResource.

src/main/java/edu/illinois/library/cantaloupe/resource/iiif/v1/InformationResource.java
jcoyne: That may be better suited for a follow up change like this https://github.com/cantaloupe-project/cantaloupe/pull/878/changes/4641a47b7969b5a5f2c4c97f79a63e0c581ebea2

DiegoPino (approved):
This looks like (feature wise) a 1:1 to what was there, but cleaner.

Approving bc it does change existing behavior, though we could eventually figure out why there is a distinction between internal and public identifier, if the later can be overridden via. a header

PR #935, opened 2026-05-01
Extract TemplateVariables class from AbstractResource

jcoyne:
Depends on #943

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/resource/ErrorResource.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/resource/ErrorResource.java
@@ -10,9 +10,8 @@ import edu.illinois.library.cantaloupe.operation.ValidationException;
 import edu.illinois.library.cantaloupe.processor.OutputFormatException;
 import edu.illinois.library.cantaloupe.processor.SourceFormatException;
 import edu.illinois.library.cantaloupe.resource.iiif.FormatException;
-import jakarta.servlet.http.HttpServletResponse;
 import jakarta.servlet.http.HttpServletRequest;
-
+import jakarta.servlet.http.HttpServletResponse;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.thymeleaf.TemplateEngine;
@@ -27,8 +26,6 @@ import java.io.StringWriter;
 import java.io.UnsupportedEncodingException;
 import java.nio.file.AccessDeniedException;
 import java.nio.file.NoSuchFileException;
-import java.util.HashMap;
-import java.util.Map;
 
 /**
  * Translates a {@link Throwable} to an HTTP 4xx or 5xx-level response.
@@ -94,8 +91,7 @@ class ErrorResource {
         final Status status = toStatus(error);
         log(status.getCode());
 
-        final Map<String,Object> templateVars = new HashMap<>();
-        templateVars.put("baseUri", request.getContextPath());
+        final TemplateVariables templateVars = new TemplateVariables();
         templateVars.put("pageTitle", status.toString());
         templateVars.put("message", error.getMessage());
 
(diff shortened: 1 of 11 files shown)


src/main/java/edu/illinois/library/cantaloupe/resource/admin/AdminResource.java
```diff
@@ -152,9 +154,9 @@ public void doGET() throws Exception {
      * @return Map containing keys that will be used as variables in the admin
      *         interface's HTML template.
      */
-    private Map<String,Object> getTemplateVars() {
-        final Map<String, Object> vars = getCommonTemplateVars();
-        vars.put("adminUri", vars.get("baseUri") + Route.ADMIN_PATH);
+    private TemplateVariables getTemplateVars() {
+        final TemplateVariables vars = TemplateVariables.getDefault(getRequest());
+        vars.put("adminUri", StringUtils.stripEnd((String) vars.get("basePath"), "/") + Route.ADMIN_PATH);
```
glenrobson, line 159: Should this be baseUri?

src/main/java/edu/illinois/library/cantaloupe/resource/TemplateVariables.java
```diff
@@ -0,0 +1,32 @@
...
+
+public class TemplateVariables {
+    public static TemplateVariables getDefault(Request request) {
+        TemplateVariables set = new TemplateVariables();
+        set.put("version", Application.getVersion());
+        set.put("basePath", request.
+                                getHeaders().
+                                getFirstValue("X-Forwarded-Path", "/"));
+        return set;
+    }
```
glenrobson, line 16: Original get common variables:
```
 String baseURI = getRequest().getPublicRootReference().toString();
            // Normalize the base URI. Note that the <base> tag will need it to
            // have a trailing slash.
            if (baseURI.endsWith("/")) {
                baseURI = baseURI.substring(0, baseURI.length() - 2);
            }
            vars.put("baseUri", baseURI);
```

glenrobson: Defined here: https://github.com/cantaloupe-project/cantaloupe/blob/53907e9041a5caf1fc02613560c03b94b6d84813/src/main/java/edu/illinois/library/cantaloupe/resource/Request.java#L128

src/main/java/edu/illinois/library/cantaloupe/resource/ErrorResource.java
```diff
@@ -94,7 +94,7 @@ public void doGET() throws Exception {
         final Status status = toStatus(error);
         log(status.getCode());
 
-        final Map<String,Object> templateVars = new HashMap<>();
+        final TemplateVariables templateVars = new TemplateVariables();
         templateVars.put("baseUri", request.getContextPath());
```
glenrobson: This won't be defined. Should we use basePath or define baseURI in the defaults?

DiegoPino: Should this
templateVars.put("baseUri", request.getContextPath());
be also named basePath and then the actual template use that instead of baseUri?

glenrobson: We may not need baseURI: https://github.com/cantaloupe-project/cantaloupe/blob/53907e9041a5caf1fc02613560c03b94b6d84813/src/main/resources/error.html

src/main/java/edu/illinois/library/cantaloupe/resource/TemplateVariables.java
glenrobson: Discussed this and we think its OK because we only need the basePath not the baseUri in the templates.

DiegoPino (approved):
Great. All looks good

PR #943, opened 2026-05-13
Use host relative paths for static resources

jcoyne:
This makes the proxy configuration simpler, as we're no longer depending on host, protocol and port for every static resource. Previously, if you were using a proxy, but didn't set X-Forwared-Port, static resources wouldn't load as it would use the Jetty port.


DiegoPino (approved):
Looks good. Thanks for updating the templates too!

PR #945, opened 2026-05-13
Update jackson

jcoyne:


pom.xml
```diff
@@ -310,7 +310,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-annotations</artifactId>
-            <version>${jackson.version}</version>
+            <version>2.21</version>
```
jcoyne, line 313: jackson-annotation no longer uses the patch level version component https://mvnrepository.com/artifact/com.fasterxml.jackson.core/jackson-annotations

glenrobson (approved):

PR #948, opened 2026-05-15
Make OperationList.applyNonEndpointMutations idempotent

### LINKED ISSUE #947: applyNonEndpointMutations runs twice per request, doubling overlays/redactions/sharpen
## Summary

`ImageRequestHandler.handle()` calls `OperationList.applyNonEndpointMutations()` twice on the same `OperationList` instance in the cache-miss path, and `applyNonEndpointMutations` is not idempotent. As a result, every cache miss appends a duplicate of each non-endpoint mutation (overlay, redaction, sharpen) to the operation list, doubling both the rendered effect and the processing cost.

## Affected versions

Reproduced on tag `v5.0.7` and on `develop` (HEAD `b812593` at time of writing). The two call sites are still at:

- `src/main/java/edu/illinois/library/cantaloupe/resource/ImageRequestHandler.java:316` (inside the info-cache lookup block, used to build the derivative cache key)
- `src/main/java/edu/illinois/library/cantaloupe/resource/ImageRequestHandler.java:405` (after `requestContext` is populated, just before `freeze()`)

## Why it duplicates

Inside `OperationList.applyNonEndpointMutations`:

```java
// Redactions
for (Redaction redaction : redactions) {
    addBefore(redaction, Encode.class);          // no presence check
}
// Sharpening
if (sharpen > 0.001) {
    addBefore(new Sharpen(sharpen), Encode.class); // no presence check
}
// Overlay
overlay.ifPresent(ov -> addBefore(ov, Encode.class)); // no presence check
```

`addBefore` blindly inserts. A second invocation appends an additional copy of each.

The `Scale` block at the top of the method is already guarded by `getFirst(Scale.class)`, and the per-encode customizations only mutate an existing `Encode`, so those parts are accidentally idempotent. The three operations above are not.

## Trigger condition

The bug fires whenever both are true:

1. The info-cache lookup at line 313 returns a present `Optional<Info>` (e.g. `cache.server.info.enabled = true`).
2. The derivative cache stream at line 320 comes back `null` (derivative cache disabled, or enabled but empty for this key).

In that path the handler runs the first `applyNonEndpointMutations` (line 316) to form the derivative cache key, falls through `cacheStream == null`, sets `isFormatKnownYet = true`, walks the format iterator, populates `requestContext`, then runs `applyNonEndpointMutations` again at line 405. Both invocations append.

Default config (info cache on, derivative cache off) hits this on every image request.

## Reproduction

Minimal config (`cantaloupe.properties`):

```
overlays.strategy = BasicStrategy
overlays.BasicStrategy.enabled = true
overlays.BasicStrategy.type = image
overlays.BasicStrategy.image = /path/to/watermark.png
overlays.BasicStrategy.position = repeat
overlays.BasicStrategy.inset = 20
overlays.BasicStrategy.output_width_threshold = 1
overlays.BasicStrategy.output_height_threshold = 1

cache.server.info.enabled = true
cache.server.derivative.enabled = false
```

Hit any image endpoint twice (first request warms the info cache; second exercises the bug). The DEBUG log line emitted by `ImageRepresentation` shows the operation list as a string, and you will see two consecutive `imageoverlay:` segments:

```
TurboJpegProcessor processed in 85 msec: <id>_1:1_cropbypixels:...,512,512_scalebypixels:512,512,box_imageoverlay:file:///.../watermark.png_REPEAT_20_imageoverlay:file:///.../watermark.png_REPEAT_20_encode:jpg_JPEG_80...
```

Switching `cache.server.info.enabled = false` makes the second `imageoverlay:` go away because the first call site is gated behind `optInfo.isPresent()`, confirming the duplication path.

## Visible effects

- Watermark renders denser than configured (alpha stacks). For a translucent repeating watermark this is visible to the eye and produces a different JPEG byte stream:
  - With bug: `29247 B`, md5 `e05a6a94a5964e73b9e21999bdfcdcbc`
  - Without bug: `28697 B`, md5 `ae23796bc0dcc3b3221dcea199c405ef`
  - Same source, same crop, same request URL.
- ~30% higher per-tile processing time on local measurements (TurboJpegProcessor, 512×512 tile, baseline JPEG output quality 80). Wall-clock benefit is bounded by the overlay/sharpen cost share of the pipeline.
- For installations using script-driven redactions the rectangles are redacted twice with full opacity each pass; for some redaction modes this is benign, for others it is not.

## Proposed fix

PR follows: make `OperationList.applyNonEndpointMutations` idempotent via a single-shot flag. This keeps the caller's two-call structure intact (the first call is still needed to populate the cache key before the derivative cache lookup) and protects future callers from the same trap. A regression test asserts that overlay, redaction, and sharpen each appear exactly once after two consecutive calls.

Alternative fixes considered:

- Track "mutationsApplied" inside `ImageRequestHandler.handle()` and skip the second call. More surgical, but does not protect against the next caller that makes the same mistake.
- Add a presence check before each `addBefore` inside `applyNonEndpointMutations`. Spreads idempotency across many small checks; the single-flag form is harder to regress.

damonsson:
Fixes #947.

## Problem

`ImageRequestHandler.handle()` calls `OperationList.applyNonEndpointMutations()` twice on the same `OperationList` instance in the cache-miss path (once at line 316 to form the derivative cache key, once at line 405 after the request context is fully populated). Several mutations inside `applyNonEndpointMutations` append operations via `addBefore` without checking whether an equivalent operation is already present:

- Redactions (one `addBefore` per redaction returned by the delegate)
- Sharpen (when `processor.sharpen > 0`)
- Overlay (when overlays are enabled)

So the second invocation appends a duplicate of each. With the Basic overlay strategy this surfaces as a visibly denser watermark and a ~30% per-tile processing penalty on every cache miss. See #947 for byte-level reproduction.

## Fix

Add a single-shot flag (`haveAppliedNonEndpointMutations`) to `OperationList`. The method is now idempotent: subsequent invocations return immediately. This keeps the existing two-call structure in `ImageRequestHandler` intact - the first call is still needed to populate the cache key before the derivative cache lookup - and protects future callers from the same trap.

I considered tracking the state at the caller (in `ImageRequestHandler.handle()`) instead, but the defensive form is harder to regress.

## Test

Added `OperationListTest#applyNonEndpointMutationsIsIdempotent`: configures Basic overlay + non-trivial sharpen + delegate redactions, calls `applyNonEndpointMutations` twice on the same list, asserts overlay/redaction/sharpen each appear exactly once.

## Test plan

- [x] `mvn -Dtest=OperationListTest test` - all 85 existing tests still pass plus the new idempotency assertion
- [x] `mvn -Dtest=ImageRequestHandlerTest test` - all 14 existing handler tests pass with no change
- [x] Manual reproduction from #947 verified: the duplicate `imageoverlay:` segment disappears from the debug operation-list log after the fix, and the rendered tile byte stream matches the single-overlay reference (`md5 ae23796bc0...`).

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/operation/OperationList.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/operation/OperationList.java
@@ -110,6 +110,7 @@ public final class OperationList implements Iterable<Operation> {
             LoggerFactory.getLogger(OperationList.class);
 
     private boolean isFrozen;
+    private boolean haveAppliedNonEndpointMutations;
     private Identifier identifier;
     private MetaIdentifier metaIdentifier;
     private final List<Operation> operations = new ArrayList<>();
@@ -221,12 +222,29 @@ public final class OperationList implements Iterable<Operation> {
      * <p>The instance's identifier must be {@link #setIdentifier(Identifier)
      * set}.</p>
      *
+     * <p>This method is idempotent: subsequent calls on the same instance are
+     * no-ops. Several of the mutations append operations unconditionally, so
+     * repeated invocations would otherwise stack duplicate redactions, sharpen
+     * passes, and overlays on the same list.</p>
+     *
      * @param info          Source image info.
      * @param delegateProxy Delegate proxy for the current request.
      */
     public void applyNonEndpointMutations(final Info info,
                                           final DelegateProxy delegateProxy) {
         checkFrozen();
+        // Several mutations below (Redaction, Sharpen, Overlay) call
+        // addBefore() without checking whether an equivalent operation is
+        // already present, so calling this method more than once on the same
+        // instance would append duplicates. ImageRequestHandler.handle()
+        // legitimately needs to call this twice in the cache-miss path (once
+        // to form the derivative cache key, once after the request context is
+        // fully populated), so guard against the second call here rather than
+        // requiring every caller to track state.
+        if (haveAppliedNonEndpointMutations) {
+            return;
+        }
+        haveAppliedNonEndpointMutations = true;
 
         // If there is a scale constraint set, but no Scale operation, add one.
         if (getScaleConstraint().hasEffect()) {
--- a/src/test/java/edu/illinois/library/cantaloupe/operation/OperationListTest.java
+++ b/src/test/java/edu/illinois/library/cantaloupe/operation/OperationListTest.java
@@ -403,6 +403,39 @@ class OperationListTest extends BaseTest {
     }
 
     @Test
+    void applyNonEndpointMutationsIsIdempotent() {
+        BasicStringOverlayServiceTest.setUpConfiguration();
+        Configuration config = Configuration.getInstance();
+        config.setProperty(Key.PROCESSOR_SHARPEN, 0.5);
+
+        final Dimension fullSize   = new Dimension(2000, 1000);
+        final Info info            = Info.builder().withSize(fullSize).build();
+        final OperationList opList = OperationList.builder()
+                .withIdentifier(new Identifier("redacted"))
+                .withOperations(new Encode(Format.get("tif")))
+                .build();
+
+        DelegateProxy proxy = TestUtil.newDelegateProxy();
+        proxy.getRequestContext().setOperationList(opList, fullSize);
+
+        opList.applyNonEndpointMutations(info, proxy);
+        opList.applyNonEndpointMutations(info, proxy);
+
+        long overlays = opList.stream()
+                .filter(op -> op instanceof Overlay).count();
+        long redactions = opList.stream()
+                .filter(op -> op instanceof Redaction).count();
+        long sharpens = opList.stream()
+                .filter(op -> op instanceof Sharpen).count();
+        assertEquals(1, overlays,
+                "overlay should be appended exactly once");
+        assertEquals(1, redactions,
+                "redaction should be appended exactly once");
+        assertEquals(1, sharpens,
+                "sharpen should be appended exactly once");
+    }
+
+    @Test
     void applyNonEndpointMutationsWithRedactions() {
         final Dimension fullSize   = new Dimension(2000, 1000);
         final Info info            = Info.builder().withSize(fullSize).build();


jcoyne:
@glenrobson @DiegoPino this is one of those cases, where having the builder pattern is making bugs like this more complicated. With the current builder pattern it's not easy to see that the OperationList is set only when the object is instantiated. https://github.com/cantaloupe-project/cantaloupe/pull/844 would resolve this. 


@damonsson thank you for raising this issue and proposing a fix. I wonder if there is some way we could refactor ImageRequestHandler to dup() the operationList in the cache lookup.  I don't really like the complexity of adding state to the OperationList, to know if a method has been called more than one time.

jcoyne:
@damonsson do you think #949 would work?

damonsson:
Thanks @jcoyne, and I agree the idempotency flag isn't elegant. Tracking "have I been called" is state I'd rather not add either.

I think #949 is the right shape, but as written it doesn't fully fix the bug. `OperationList.copy()` there is a shallow copy:

```java
copy.operations.addAll(this.operations);
```

It copies the list, but the `Operation` objects inside are shared by reference between `cacheOpList` and the original `operationList`. `applyNonEndpointMutations()` does two kinds of mutation:

1. **Appends to the list** (Overlay, Redaction, Sharpen via `addBefore`). #949 handles these correctly, since they go onto the copy's own list.
2. **Mutates pre-existing operations in place**: `crop.setOrientation()`, `scale.setLinear()`/`setFilter()`, `encode.setCompression()`/`setInterlacing()`/`setQuality()`. These run against the shared objects.

Most of those are idempotent setters, so applying them twice is harmless. The exception is `rotate.addDegrees(orientation.getDegrees())`, which is additive. In the cache-miss path `cacheOpList.applyNonEndpointMutations()` rotates the shared `Rotate`, then the later `operationList.applyNonEndpointMutations()` rotates the same instance again, so the orientation is applied twice for any source image whose embedded metadata declares a non-zero orientation.

So #949 fixes the visible symptom (the doubled overlay) but leaves the same class of bug for embedded-orientation rotation.

For #949 to be fully correct, `copy()` would need to be a deep copy that also clones the `Operation` instances. Today `Operation` has no copy/clone contract, so that is a bigger change than it looks: it would mean adding something like an `Operation copy()` method to the interface and every implementation.

Given that, two honest options:

- Add an `Operation.copy()` contract and make `OperationList.copy()` deep. Cleanest long-term, makes #949 fully correct and keeps `OperationList` stateless.
- Or guard `applyNonEndpointMutations` against running twice (#948), accepting the small bit of state.

I'm happy to go whichever way you prefer. I can update #948 toward the deep-copy approach, or close it in favour of a corrected #949. Just let me know.

jcoyne:
@damonsson great. If you're interested in looking at the deep copy approach, please go ahead. I think that approach is the way to go. Thanks again for looking into this.

jcoyne:
Superseded by #949

PR #949, opened 2026-05-15
Copy the operationsList before modifying it

jcoyne:
See #948

OperationList.applyNonEndpointMutations() twice on the same OperationList instance in the cache-miss path (once at line 316 to form the derivative cache key, once at line 405 after the request context is fully populated). Several mutations inside applyNonEndpointMutations append operations via addBefore without checking whether an equivalent operation is already present

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/operation/ColorTransform.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/operation/ColorTransform.java
@@ -13,6 +13,14 @@ public enum ColorTransform implements Operation {
     BITONAL, GRAY;
 
     /**
+     * @return The instance itself, as it is immutable.
+     */
+    @Override
+    public Operation copy() {
+        return this;
+    }
+
+    /**
      * Does nothing.
      */
     @Override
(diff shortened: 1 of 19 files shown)


glenrobson:
Note about deep copy from this one: https://github.com/cantaloupe-project/cantaloupe/pull/948#issuecomment-4458853446

DiegoPino:
@jcoyne @glenrobson maybe OpList class could `implement Cloneable` and then override the super clone method, thus implementing its own deep clone? Might save a few lines of code/remove the need for a helper method?

jcoyne:
@DiegoPino I've implemented copy() on the full list. I decided not to `implement Cloneable` after reading https://www.artima.com/articles/josh-bloch-on-design (Bloch lead the design of the Java Collections Framework).

DiegoPino:
@jcoyne all good. I just read the article you shared and it makes sense. Thanks

DiegoPino (commented):
This looks good to me, but will let @glenrobson chime in before approving

DiegoPino (approved):
Looks good! Approving

PR #954, opened 2026-06-15
Scale: don't reject !w,h when source is smaller than the requested box

### LINKED ISSUE #953: ScaleRestrictedException when specifying max size (not exact size) with `!`
## Steps to reproduce

- Using Cantaloupe 5.0.6
- Given an image of size < 600x600
- Request the image, specifying a size of `!600,600`, e.g.

  ```none
  GET https://cantaloupe/iiif/2/<id>/full/!600,600/0/default.jpg
  ```

## Expected

- Image request succeeds
- As per the [IIIF docs](https://iiif.io/api/image/2.0/#size):

  > The image content is scaled for the best fit such that the resulting width and height are less than or equal to the requested width and height. 

## Actual:

- Image request fails with `403 Forbidden`
- Response gives the stack trace:

  ```none
  edu.illinois.library.cantaloupe.resource.ScaleRestrictedException: Requests for scales in excess of 100% are not allowed.
	at edu.illinois.library.cantaloupe.resource.PublicResource.validateScale(PublicResource.java:261)
	at edu.illinois.library.cantaloupe.resource.iiif.v2.ImageResource.access$300(ImageResource.java:31)
	at edu.illinois.library.cantaloupe.resource.iiif.v2.ImageResource$1CustomCallback.willProcessImage(ImageResource.java:112)
	at edu.illinois.library.cantaloupe.resource.ImageRequestHandler.handle(ImageRequestHandler.java:403)
	at edu.illinois.library.cantaloupe.resource.iiif.v2.ImageResource.doGET(ImageResource.java:128)
	at edu.illinois.library.cantaloupe.resource.HandlerServlet.handle(HandlerServlet.java:97)
	at edu.illinois.library.cantaloupe.resource.HandlerServlet.doGet(HandlerServlet.java:35)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:645)
  ```

## Related issues

- #600

avalyset:
Fixes #953.

A IIIF `!w,h` request against a source smaller than the box returns
403 ScaleRestrictedException under the default `max_scale = 1.0`,
instead of returning the source at its own size.

## Reproduce

Source 300×300, `max_scale = 1.0` (sample default):

```
GET .../full/!600,600/0/default.jpg → 403 Forbidden (expected: 200, 300×300)
```

## Why this is wrong

IIIF Image API 3.0 §4.2, for `!w,h`: the returned image "must be as
large as possible but not larger than the extracted region, w or h".
For a 300×300 source and `!600,600`, the correct result is 300×300 —
no upscaling. v3 uses the separate `^!w,h` (caret) syntax when
upscaling *is* wanted. v2 §4.2 allows the server to decline upscaling,
which `max_scale = 1.0` does.

The bug: `ScaleByPixels` computes a scale factor of 2.0 for the
`ASPECT_FIT_INSIDE` case regardless of whether upscaling is permitted,
so the validator rejects a request that should simply return the
source size.

## Fix

Carry whether upscaling is allowed through to the scale computation,
rather than fixing it only at validation (which would still upscale on
the render path) or clamping unconditionally (which would break the
legitimate `^!w,h` upscale path):

- `ScaleByPixels` gains an `upscaleAllowed` flag (defaults to `true`,
  so the direct Java API and all existing callers are unchanged). When
  false, `ASPECT_FIT_INSIDE` clamps the scale to ≤ 1.0.
- v2 `Size.toScale()` sets it `false` for `!w,h` (v2 has no caret).
- v3 `Size.toScale()` propagates the parser's existing
  `isUpscalingAllowed()` — `false` for `!w,h`, `true` for `^!w,h`.

`equals`/`hashCode` updated to include the flag so cached operations
stay correct.

## Tests

- v2 and v3 `!600,600` on a 300×300 source → 300×300, no exception.
- v3 `^!600,600` with `max_scale = 2.0` → 600×600 (upscale path intact).
- `pct:` scaling unaffected.
- All existing ScaleByPixels / Size tests pass (default `upscaleAllowed
  = true` path unchanged).

One related observation, left out of this change: `ScaleValidator` and
the `v3` resource default `max_scale` to 1.0, while `InformationFactory`
advertises an unbounded `maxScale` in info.json. Happy to open a
separate issue if useful.

(Verified against the expected outputs @dmolesUC listed in the issue: 150×75 and 1200×150 sources both return at source size under max_scale = 1.0.)

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/operation/ScaleByPixels.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/operation/ScaleByPixels.java
@@ -41,6 +41,16 @@ public class ScaleByPixels extends Scale implements Operation {
     private Mode scaleMode;
 
     /**
+     * Whether the operation is permitted to scale the source above its
+     * original dimensions. Defaults to {@code true} for backward compatibility
+     * with direct API callers. IIIF callers set this from the URI: v2 {@code !w,h}
+     * is always {@code false}; v3 {@code !w,h} is {@code false}, {@code ^!w,h}
+     * is {@code true}. When {@code false}, {@link Mode#ASPECT_FIT_INSIDE}
+     * clamps its resulting scale to 1.0 so the source is never upscaled.
+     */
+    private boolean upscaleAllowed = true;
+
+    /**
      * No-op constructor.
      */
     public ScaleByPixels() {
@@ -77,7 +87,8 @@ public class ScaleByPixels extends Scale implements Operation {
             return Objects.equals(other.getWidth(), getWidth()) &&
                     Objects.equals(other.getHeight(), getHeight()) &&
                     Objects.equals(other.getMode(), getMode()) &&
-                    Objects.equals(other.getFilter(), getFilter());
+                    Objects.equals(other.getFilter(), getFilter()) &&
+                    other.isUpscaleAllowed() == isUpscaleAllowed();
         }
         return super.equals(obj);
     }
@@ -95,6 +106,28 @@ public class ScaleByPixels extends Scale implements Operation {
 
     public Mode getMode() {
         return scaleMode;
+    }
+
+    /**
+     * @return Whether the instance is permitted to scale above the source
+     *         dimensions when in {@link Mode#ASPECT_FIT_INSIDE} mode.
+     *         Defaults to {@code true} unless explicitly disabled.
+     */
+    public boolean isUpscaleAllowed() {
+        return upscaleAllowed;
+    }
+
+    /**
+     * @param upscaleAllowed Whether the instance may scale above source
+     *                       dimensions. IIIF v2 {@code !w,h} and IIIF v3
+     *                       {@code !w,h} (no caret) must set this to
+     *                       {@code false}; v3 {@code ^!w,h} sets it to
+     *                       {@code true}.
+     * @throws IllegalStateException if the instance is frozen.
+     */
+    public void setUpscaleAllowed(boolean upscaleAllowed) {
+        checkFrozen();
+        this.upscaleAllowed = upscaleAllowed;
     }
 
     /**
@@ -119,6 +152,9 @@ public class ScaleByPixels extends Scale implements Operation {
                 double xScale = getWidth() / reducedSize.width();
                 double yScale = getHeight() / reducedSize.height();
                 rfScale       = Math.min(xScale, yScale);
+                if (!upscaleAllowed) {
+                    rfScale = Math.min(rfScale, 1.0);
+                }
                 break;
         }
         ReductionFactor rf = ReductionFactor.forScale(rfScale);
@@ -141,9 +177,13 @@ public class ScaleByPixels extends Scale implements Operation {
                 result[0] = result[1] = getWidth() / fullSize.width();
                 break;
             case ASPECT_FIT_INSIDE:
-                result[0] = result[1] = Math.min(
+                double insideScale = Math.min(
                         getWidth() / fullSize.width(),
                         getHeight() / fullSize.height());
+                if (!upscaleAllowed) {
+                    insideScale = Math.min(insideScale, 1.0);
+                }
+                result[0] = result[1] = insideScale;
                 break;
             default:
                 result[0] = getWidth() / fullSize.width();
@@ -184,6 +224,9 @@ public class ScaleByPixels extends Scale implements Operation {
                 scalePct = Math.min(
                         getWidth() / size.width(),
                         getHeight() / size.height());
+                if (!upscaleAllowed) {
+                    scalePct = Math.min(scalePct, 1.0);
+                }
                 size.setWidth(size.width() * scalePct);
                 size.setHeight(size.height() * scalePct);
                 break;
@@ -238,11 +281,12 @@ public class ScaleByPixels extends Scale implements Operation {
 
     @Override
     public int hashCode() {
-        int[] codes = new int[4];
+        int[] codes = new int[5];
         codes[0] = getWidth();
         codes[1] = getHeight();
         codes[2] = (getMode() != null) ? getMode().hashCode() : 0;
         codes[3] = (getFilter() != null) ? getFilter().hashCode() : 0;
+        codes[4] = Boolean.hashCode(isUpscaleAllowed());
         return Arrays.hashCode(codes);
     }
 
--- a/src/main/java/edu/illinois/library/cantaloupe/resource/iiif/v2/Size.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/resource/iiif/v2/Size.java
@@ -184,8 +184,15 @@ class Size {
                 return new ScaleByPixels(
                         null, getHeight(), ScaleByPixels.Mode.ASPECT_FIT_HEIGHT);
             case ASPECT_FIT_INSIDE:
-                return new ScaleByPixels(
+                ScaleByPixels insideScale = new ScaleByPixels(
                         getWidth(), getHeight(), ScaleByPixels.Mode.ASPECT_FIT_INSIDE);
+                // IIIF Image API v2 §4.2: !w,h scales to be "less than or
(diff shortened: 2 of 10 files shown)


glenrobson:
Glen to check and merge

avalyset:
Heads-up on the scope of this before merge, Glen.

The v3 CI failure (Version3_0ConformanceTest.testSizeDownscaledToFitInsideWithIllegalSize) surfaced that the original patch over-reached: it extended the #953 relaxation into IIIF v3 as well as v2.

That was wrong for v3. In v3, a plain !w,h size that would require upscaling must still be rejected with 400 — ^!w,h is the explicit v3 opt-in for upscaling. That rejection is correct spec behaviour and needs to stay. #953 itself was reported against v2 (max_scale = 1.0), where there is no caret and the server declining to upscale is the right outcome.

Latest push limits the relaxation to v2:
- reverts the v3 propagation in v3/Size.java, and the two v3 SizeTest assertions I'd wrongly added;
- keeps the v2 fix and the ScaleByPixels upscaleAllowed flag untouched.

Version3_0ConformanceTest is green again. Full local run: 279 tests, 0 failures (JDK 21).

The other two red jobs in the previous run were infra, not this diff — OpenJDK 26 tarball wget failing in the Dockerfile, and the Windows docker daemon not being up.

avalyset:
The latest commit is empty — it only retriggers CI, no code change.

The earlier red was one matrix cell out of sixteen: `build (ubuntu-22.04, jdk21)`, failing on `S3CacheTest.testPurgeWithIdentifier`, an S3/seaweedfs integration test. The same code passed on jdk25, jdk26, and graalvm on the same OS, so this reads as flaky infrastructure rather than the IIIF `!w,h` scaling logic in this PR.

The re-run needs maintainer approval to start (fork PR). Let me know if you'd like anything changed.

HeineDeelstra:
> That was wrong for v3. In v3, a plain !w,h size that would require upscaling must still be rejected with 400 — ^!w,h is the explicit v3 opt-in for upscaling. That rejection is correct spec behaviour and needs to stay.

Would it _require_ upscaling? Doesn't the 3.0 spec _technically_ allow a scale factor of 1.0 for an image smaller than !w,h ("as large as possible but not larger than the extracted region, w or h, or ..."?

dmolesUC:
The [3.0 spec](https://iiif.io/api/image/3.0/#42-size) does say

> Requests for sizes not prefixed with ^ that result in a scaled region with pixel dimensions greater than the pixel dimensions of the extracted region are errors that should result in a 400 (Bad Request) status code

but I don't think a `!w,h` falls into that category -- as @HeineDeelstra points out, for `!w,h` the spec says 

> scaled so that the width and height of the returned image are not greater than `w` and `h`, while maintaining the aspect ratio. The returned image must be as large as possible but not larger than the extracted region, `w` or `h`, or server-imposed limits

That says to me that `!w,h` *can't* result in an invalid scaled region; it's defined to return the largest *valid* scaled region less than or equal to `w × h` (with the original aspect ratio).

avalyset:
My July note was wrong.

`!w,h` carries "not larger than the extracted region" inside its own definition, so its scale factor is min(w/rw, h/rh, 1) by construction. It can't produce a scaled region larger than the source, so the 400 clause for non-caret sizes has nothing to bite on. `^!w,h` drops "the extracted region" from that same sentence, which is what makes it the upscaling form.

Two commits from Aug 7 put the branch back on that reading:

- `37ea954` restores the v3 propagation, so plain `!w,h` gets `upscaleAllowed=false` while `^!w,h` keeps it true.
- `8c4f5b0` adds v2 end-to-end coverage — #953 was reported against a v2 URL, but `!w,h` had only unit-level tests on that side.

One of those needs your sign-off rather than mine. `Version3_0ConformanceTest.testSizeDownscaledToFitInsideWithIllegalSize` quotes the "not larger than the extracted region" passage in its own javadoc, and then asserts a 400 for `!300,300` against the 64x56 test image. The test contradicts the text it cites. I changed it to expect 64x56 and renamed it accordingly — but that's a conformance test changing its mind about what conformance is, so if either of you reads the spec differently, I'd rather back it out than have it land quietly.

Scope note: only `ASPECT_FIT_INSIDE` is affected. `w,h`, `w,`, `,h` and `pct:` keep the default, so the v3 `ImageResource` guard still returns 400 for those when they would scale above 100%.

CI hasn't run on the current head — the fork run is sitting at `action_required`. Could one of you approve it?

glenrobson:
Cantaloupe dev team looked at this today. We think that !w,h should return the closest matching image to w,h without upscaling for v2 and v3. For example if it is a 300,300 image:

!600,600 -> 300 by 300 pixel image
^!600,600 -> 600 by 600 upscaled image

/0,0,150,150/!600,600 -> 150,150
/0,0,150,150/^!600,600 -> 600 by 600 

If server doesn't support upscailing then return 501

I've set the tests to run and will look at the implementation.

avalyset:
Implemented in 507007ba482a399ccbc3059cedb7c97556126a06.

The boundary I used for "does not support upscaling" is `max_scale`, because the
codebase already draws it there: `InformationFactory` withholds `sizeUpscaling` from
the v3 information response when `max_scale` is 1.0 or lower
(`resource/iiif/v3/InformationFactory.java:142`). A caret size that needs upscaling
under such a ceiling now returns 501, while the same request under a ceiling above 1.0
— where upscaling is supported and advertised, and the request merely overshoots —
stays 400. The two part ways only at `max_scale = 0`, which means no ceiling: nothing
is rejected on this path there, so the status choice never surfaces.

Applies to `^w,`, `^,h`, `^pct:n`, `^w,h` and `^!w,h`. `^max` is unaffected: `v3/Size.java`
clamps it to the ceiling, so it never reaches the rejection path. One narrowing: a
ceiling below 1.0 also rejects sizes that scale down, such as `^pct:60` under
`max_scale = 0.5`. Those stay 400 — 501 is reserved for requests that actually need
upscaling.

No behaviour changes in v1 or v2. Neither has caret syntax, so the 501 clause in 4.2 has
nothing to bite on there, and their `max_scale` rejections remain 403.

Five javadoc notes in `Version3_0ConformanceTest` said 400 was chosen because upscaling
is a continuum rather than a switch. I rewrote them rather than deleting them — they now
say where the boundary is and why, and point at the new test that keeps the 400 side
covered.

The affected suites go from 238 to 241 tests, 489 across the v1/v2/v3 resource suites,
0 failures on JDK 21.

On your run of the previous head: the one red cell, `build (windows-latest, jdk21)`,
failed before compiling — the Windows docker daemon wasn't up. `windows-latest/jdk25`
built and tested the same merge commit and passed.

One thing I left out of this change, happy to file separately: scale validation is
full-image-scoped while rendering is region-scoped. `/0,0,32,28/^!128,112/` under
`max_scale = 2.0` serves a 4x upscale of the region, because 128/64 measures 2x against
the full image. Your third and fourth cases already behave correctly; this is the
permissive direction of the same asymmetry.

avalyset:
Your four cases, checked against the branch.

The first two were already covered by the suite: `!300,300` against the 64×56 source returns 64×56, and `^!100,100` upscales to 100×88. The 501 for a caret size without server support is in `507007ba`. The two region-scoped cases had no test, so I added them in `ca1e70c8`:

- `/0,0,32,28/!600,600` → 32×28. The ceiling in the `!w,h` definition is the extracted region, not the full image, so the box is clamped to the region.
- `/0,0,32,28/^!600,600` → 600×525. The caret form drops the region from that ceiling. It is 525 rather than 600 because the region's 8:7 aspect is preserved.

Both pass. `Version3_0ConformanceTest` is 57 tests, 0 failures on JDK 21.

That covers everything in your 19 August note. The one thing still outstanding is the asymmetry I mentioned then: scale validation is full-image-scoped while rendering is region-scoped, so `/0,0,32,28/^!128,112/` under `max_scale = 2.0` serves a 4× upscale of the region because 128/64 measures 2× against the full image. Your cases behave correctly; this is the permissive direction of the same measurement. Happy to file it separately if you want it addressed.

PR #958, opened 2026-07-06
Fix ArrayIndexOutOfBoundsException scaling grayscale+alpha images

rogerhoward:
## Problem

Applying a **scaled grayscale+alpha PNG** overlay (PNG color type 4) throws `ArrayIndexOutOfBoundsException` in the Java2D overlay compositor, returning a 500.

The failure is not in *reading* the PNG — ImageIO decodes it fine into a 2-band `TYPE_CUSTOM` image. It's in the **scaling** step: `Java2DUtil.scale()` runs the overlay through the custom `ResampleOp` resampler, which only handles 1-, 3-, and 4-channel images. For a 2-channel image the sampling loops still read three samples per pixel out of a `srcWidth * 2` row, running off the end:

```
ArrayIndexOutOfBoundsException: Index 200 out of bounds for length 200
    at ResampleOp.horizontalFromSrcToWork(ResampleOp.java:419)
    at ResampleOp.doFilter(ResampleOp.java:259)
```

## Fix

Extend the existing up-front normalization in `ResampleOp.doFilter` — which already converts binary/indexed images — to also convert 2-band (gray+alpha) images to `TYPE_4BYTE_ABGR`, preserving alpha. This fixes it at the root, covering both scaled overlays and gray+alpha source images that get scaled.

## Testing

- Added regression test `Java2DUtilTest.scaleWithGrayAlphaImage`.
- Verified against the real `ResampleOp` source: reproduces the AIOOBE before the change; scales cleanly to `TYPE_4BYTE_ABGR` after.

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/processor/resample/ResampleOp.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/processor/resample/ResampleOp.java
@@ -221,8 +221,15 @@ public class ResampleOp extends AdvancedResizeOp {
                     " but must be at least 3x3.");
         }
 
+        // This resampler only supports 1- (gray), 3- (RGB), and 4-channel
+        // (ARGB) images. Two-channel gray+alpha images (e.g. PNG color type 4)
+        // and palette/binary images would otherwise cause the horizontal and
+        // vertical sampling loops to read three or four channels out of a
+        // narrower row, throwing ArrayIndexOutOfBoundsException. Convert them
+        // up front to a supported layout, preserving alpha when present.
         if (srcImage.getType() == BufferedImage.TYPE_BYTE_BINARY ||
-                srcImage.getType() == BufferedImage.TYPE_BYTE_INDEXED) {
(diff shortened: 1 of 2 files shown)


rogerhoward:
Hi! I'm not sure if you accept PRs - I've tested this patch on our end as best as I can; we implemented a workaround in production, but it had us stumped for a hot minute. In short

- we had RGBA PNGs we'd sometimes process into new PNGs using Imagemagick
- turns out Imagemagick's encoder would see RGBA PNGs that had only gray values, and would quietly optimize them by converting to grayscale+alpha PNGs
- these grayscale+alpha PNGs could not be used as overlays in Cantaloupe

DiegoPino:
@rogerhoward the code change looks clean, thank you. I will let the workflows run first to see if all tests are passing.

DiegoPino:
@jcoyne @glenrobson this looks good to me, a very simple fix. Any concerns?

jcoyne (approved):

DiegoPino:
Thanks @jcoyne for approving and @rogerhoward for the fix. merged.

PR #961, opened 2026-07-08
Extract a IIIFRequest object from Request

jcoyne:
This will support things only in IIIFRequests (like identifiers).  This also extracts a RouteEntry that matches a route to the resource being served.  Certain routes (Images & Info) get IIIFRequests, while other routes get plain Requests.

### DIFF
--- a/src/main/java/edu/illinois/library/cantaloupe/resource/Route.java
+++ b/src/main/java/edu/illinois/library/cantaloupe/resource/Route.java
@@ -30,72 +30,92 @@ public final class Route {
      * N.B.: the {@link LinkedHashMap} preserves order as each mapping will be
      * checked sequentially and the first match used.
      */
-    private static final Map<Pattern,Class<? extends AbstractResource>> MAPPINGS =
+    private static final Map<Pattern,RouteEntry> MAPPINGS =
             new LinkedHashMap<>();
 
+    private static class RouteEntry {
+        Class<? extends AbstractResource> resource;
+        Class<? extends Request> request;
+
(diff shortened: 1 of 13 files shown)


src/main/java/edu/illinois/library/cantaloupe/resource/Route.java
```diff
@@ -108,8 +128,8 @@ static Route forPath(String path) {
             final Pattern pattern = entry.getKey();
             final Matcher matcher = pattern.matcher(path);
             if (matcher.find()) {
-                final Route route = new Route();
-                route.setResource(entry.getValue());
+                RouteEntry routeEntry = entry.getValue();
+                final Route route = new Route(routeEntry.getResourceClass(), routeEntry.getRequestClass());
                 for (int i = 1; i <= matcher.groupCount(); i++) {
                     route.getPathArguments().add(matcher.group(i));
                 }
```
glenrobson, line 135: Where is .getPathArguments() in the Route class below?

src/main/java/edu/illinois/library/cantaloupe/resource/Route.java
```diff
@@ -130,15 +150,22 @@ List<String> getPathArguments() {
         return pathArguments;
     }
```
glenrobson, line 157: Should this constructor be further up in the class?

jcoyne: Sure, I moved it up to just after the static methods.

glenrobson (approved):

PR #972, opened 2026-09-15
Upgrade from JDK26 to JDK27

jcoyne:
Both of these are short term java releases


DiegoPino (approved):
Looks Good, but let's wait for test to pass before merging!

